PR #11, opened 2021-10-20
Py3.9

local-minimum:
- Running py2to3
- Fixing obvious style and lint errors while inspecting changes (only `file` lint error remain, will deal with separately)
  - Line length and added `E501` where no reasonable breaking
  - Indentations
  - added `Exception` as caught when naked `except`
  - removed unused variables
  - added `E712` when comparing array to `True`/`False` 
- Type hinting added where obvious or previous comments specified types. 
- Marking with comments some changes that falls outside above
- Last commit dealt with all remaining lint errors after migrating to 3 and fixing as I browsed through the files.
- On purpose allowing docker image to fail since things aren't completed for python3 and better do that separately.

### DIFF
--- a/scanomatic/server/scanning_effector.py
+++ b/scanomatic/server/scanning_effector.py
@@ -65,7 +65,7 @@ Scan-o-Matic"""
 SOM_MAIL_BODY_SCAN_ERROR = (
     """This is an automated email, please don't reply!
 
-The project '{project_name}' on ''"""
+The project '{project_name}' on '"""
     + AppConfig().computer_human_name +
     """' reports an error while scanning.
 Please hurry to correct this so that the project won't be spoiled.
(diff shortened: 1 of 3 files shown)


scanomatic/models/factories/analysis_factories.py
```diff
@@ -142,167 +142,144 @@ def all_keys_valid(cls, keys):
...
-
-            if plate_exists and (
-                    0 <= val < dim_max for val, dim_max in
-                    zip(
+                0 <= model.focus_position[0] < len(model.pinning_matrices)
+                and model.pinning_matrices[model.focus_position[0]] is not None
+            )
+
+            if (
+                plate_exists and all(
```
local-minimum, line 218: added `all` here because:

1. A generator is always truthy (only had generator before)
2. `_validate_grid_images` a bit further down uses similar logic and has `all`.

PR #68, opened 2021-11-08
rename specs tests

### LINKED ISSUE #63: Rename specs as tests
# 📋 Info

`spec(s)` is an old Jasmine name for tests. This is confusing. We should call the files tests instead.

# 🏁 DoD

- [ ] rename files `*.spec.js*`->`*.test.js*`
- [ ] rename directory `specs`->`tests` (or `test`)
- [ ] update path in `package.json` etc. so that linting and tests work.

skymandr:
This renames the "specs" to "tests".

Also fixes a few warnings and  errors that occurred when running the tests, but which don't fail the pipeline, so please check the files.

Resolves #63


skymandr:
So, the remaining problem is with `error` in `<PolynomialConstructionContainer/>`, which should be a `string` according to the proptypes, but is set from


```javascript
  handleConstruction() {
    const { id, accessToken } = this.props.cccMetadata;
    const { degreeOfPolynomial } = this.state;
    return API.SetNewCalibrationPolynomial(id, degreeOfPolynomial, accessToken)
      .then(this.handleConstructionResults)
      .catch(this.handleConstructionResultsError);
  }

  handleConstructionResultsError(reason) {
    this.setState({ error: reason });
  }
```

Letting it be a string in the tests provokes the linter, but this could be disabled. Not sure it is a good thing that it assumes that it is a string though, so I suggest:

```javascript
  handleConstructionResultsError(reason) {
    this.setState({ error: reason.message ?? reason });
  }
```
or similar.

What do you say, @local-minimum ?

local-minimum:
> So, the remaining problem is with `error` in `<PolynomialConstructionContainer/>`, which should be a `string` according to the proptypes, but is set from
> 
> ```js
>   handleConstruction() {
>     const { id, accessToken } = this.props.cccMetadata;
>     const { degreeOfPolynomial } = this.state;
>     return API.SetNewCalibrationPolynomial(id, degreeOfPolynomial, accessToken)
>       .then(this.handleConstructionResults)
>       .catch(this.handleConstructionResultsError);
>   }
> 
>   handleConstructionResultsError(reason) {
>     this.setState({ error: reason });
>   }
> ```
> 
> Letting it be a string in the tests provokes the linter, but this could be disabled. Not sure it is a good thing that it assumes that it is a string though, so I suggest:
> 
> ```js
>   handleConstructionResultsError(reason) {
>     this.setState({ error: reason.message ?? reason });
>   }
> ```
> 
> or similar.
> 
> What do you say, @local-minimum ?

Hard to say without the lint error when fixed. Reasonably reason can be a `string`, `error` or `undefined` in the most general case... so it would be `string | undefined` after your suggestion.

If it's used as a string in the actual component then yes fix it, but attempt to fix the lint issue too rather than ignore

scanomatic/ui_server_data/js/ccc/components/PolynomialResultsColonyHistogram.jsx
```diff
@@ -106,10 +106,18 @@ export default class PolynomialResultsColonyHistogram extends React.Component {
...
         />
-        <span className="poly-colony-txt">{valueFormatter(independentMeasurement, 2)} cells</span>
+        <span className="poly-colony-txt">{valueFormatter(independentMeasurements, 2)} cells</span>
       </div>
     );
   }
 }
 
-PolynomialResultsColonyHistogram.propTypes = CCCPropTypes.colonyShape;
+PolynomialResultsColonyHistogram.propTypes = {
```
local-minimum, line 115: was it only used here or why this change?

scanomatic/ui_server_data/js/ccc/containers/PolynomialConstructionContainer.jsx
```diff
@@ -20,10 +20,8 @@ export default class PolynomialConstructionContainer extends React.Component {
     };
 
     this.handleConstruction = this.handleConstruction.bind(this);
-    this.handleConstructionResults = this.handleConstructionResults
-      .bind(this);
-    this.handleConstructionResultsError =
-            this.handleConstructionResultsError.bind(this);
+    this.handleConstructionResults = this.handleConstructionResults.bind(this);
```
local-minimum, line 23: the functions ought not to be bound here but instead be arrow functions on the class, but that's probably better a separate issue.

local-minimum (changes requested):

scanomatic/ui_server_data/js/ccc/containers/PolynomialConstructionContainer.jsx
skymandr: Sure, I don't understand the problem though, so I'd appreciate if you create it or at least commit to fleshing it out if I do.

scanomatic/ui_server_data/js/ccc/components/PolynomialResultsColonyHistogram.jsx
skymandr: This is a reversal of a change from a previous PR, which didn't quite work (gave the weird error ```ERROR: 'Warning: Failed prop type: The prop `isRequired` is marked as required in `PolynomialResultsColonyHistogram`, but its value is `undefined`.
    in PolynomialResultsColonyHistogram'``` even though the linter didn't complain). The `colonyShape` was only used here anyway.

local-minimum: then it must be removed yes.

scanomatic/ui_server_data/js/ccc/containers/PolynomialConstructionContainer.jsx
local-minimum: it's annoying that the way `this` works in those is redefined here in the constructor instead of them being defined as intended from the beginning. It has to do with javascript prototypes, so if interested google about the differences between function and arrow statement and how this affects `this`.

scanomatic/ui_server_data/js/ccc/components/PolynomialResultsColonyHistogram.jsx
skymandr: done

skymandr:
The remaining annoyance is
```console
SUMMARY:
✔ 394 tests completed
ℹ 1 test skipped
```
when running tests. For the life of me I cannot work out what test is skipped...

scanomatic/ui_server_data/js/ccc/containers/PolynomialConstructionContainer.jsx
skymandr: I am not really that interested in anything except to know whether you will create the issue or if I should. :)

local-minimum: i can make the issue

local-minimum:
> The remaining annoyance is
> 
> ```
> SUMMARY:
> ✔ 394 tests completed
> ℹ 1 test skipped
> ```
> 
> when running `npm test`. For the life of me I cannot work out what test is skipped...

it's probably something in the test runners config that sets this... no clue what though...

local-minimum (approved):

PR #75, opened 2021-11-10
Issue 52 repair tests continued 2

### LINKED ISSUE #52: STORY: Repair tests
# 📋 Info

Tests run but produce errors.

Currently in the broken state the tests create a file `./analysis/analysis.log` that should at the very least be cleaned up by the tests.

# 🏁 DoD
- [x] #76
- [x] #77
- [x] #78
- [x] #36
- [x] #37

skymandr:
This PR makes sure the integration tests run and that they create their output directory in a temporary directory. It also fixes a few additional runtime warnings.

Though they are closer to passing, all three test-suites for backend still fail, but I suggest that we close #52 and open specific tasks for each instead.


local-minimum (approved):
I agree to making 3 tasks, but instead of closing #52 lets make it a tracking task

PR #81, opened 2021-11-11
Fix Dict, Tuple and List

skymandr:
Fixes so that we use lowercase types instead of `Dict`, `Tuple` and `List`, as well as a few `dict()` etc. initialisations.

I also started looking at the numerous instances where we use `dict(a=foo, b=bar)` syntax rather than `{"a": foo, "b": bar}`, but there is a butt-load of those, and in many instances it may be justified.

I'd be happy to revert the places where I did implement this, if we agree that we prefer it this way.

### DIFF
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -1,8 +1,8 @@
 import re
-from collections import namedtuple
+from collections import namedtuple, Callable
 from enum import Enum
 from logging import Logger
-from typing import Any, Callable, Union
+from typing import Any, Union
 from uuid import uuid1
 
 import numpy as np
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -2,7 +2,7 @@ import re
 from collections import namedtuple
 from enum import Enum
 from logging import Logger
-from typing import Any, Callable, Dict, Union
+from typing import Any, Callable, Union
 from uuid import uuid1
 
 import numpy as np
@@ -350,11 +350,11 @@ def get_local_fixture_for_image(identifier, image_identifier):
     issues = {}
     fixture.set_current_areas(issues)
 
-    return dict(
-        plates=current_settings.model.plates,
-        grayscale=current_settings.model.grayscale,
-        issues=issues,
-    )
+    return {
+        "plates": current_settings.model.plates,
+        "grayscale": current_settings.model.grayscale,
+        "issues": issues,
+    }
 
 
 @_validate_ccc_edit_request
@@ -714,7 +714,7 @@ def _get_all_grid_shapes(ccc):
 
 
 @_validate_ccc_edit_request
-def construct_polynomial(identifier, power) -> Union[bool, Dict[str, Any]]:
+def construct_polynomial(identifier, power) -> Union[bool, dict[str, Any]]:
     ccc = __CCC[identifier]
     data_store = _collect_all_included_data(ccc)
     try:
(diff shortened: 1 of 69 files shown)


scanomatic/data_processing/strain_handler.py
```diff
@@ -38,10 +38,10 @@ def loadCSV2Numpy(
             will be set to nan, else they will be zero of that data type
     """
 
-    def _putInData(d, pos, m):
+    def _putInData(d: dict[tuple[int, int], list], pos: SupportsInt, m):
```
local-minimum, line 41:
```suggestion
    def _putInData(d: dict[tuple[int, int], list], pos: Sequence[SupportsInt], m):
```
or something right?

scanomatic/data_processing/strain_handler.py
```diff
@@ -1,5 +1,5 @@
 from logging import Logger
-from typing import Any, List, Union, Tuple
+from typing import Any, SupportsInt, Union
```
local-minimum, line 2: Sequence e deprecated, ska importeras från `collections.abc` enligt docs

scanomatic/data_processing/strain_selector.py
```diff
@@ -1,4 +1,4 @@
-from typing import List, Sequence, Tuple
+from typing import Sequence
```
local-minimum, line 1:
```suggestion
from collections.abs import Sequence
```

scanomatic/io/image_loading.py
```diff
@@ -1,7 +1,7 @@
 import glob
 import os
 from logging import Logger
-from typing import Optional, Sequence, Tuple
+from typing import Optional, Sequence
```
local-minimum, line 4: sequence from `collections.abs`

scanomatic/data_processing/calibration.py
```diff
@@ -2,7 +2,7 @@
 from collections import namedtuple
 from enum import Enum
 from logging import Logger
-from typing import Any, Callable, Dict, Union
+from typing import Any, Callable, Union
```
local-minimum, line 5: callable from `collections.abc`

scanomatic/image_analysis/signal.py
```diff
@@ -98,37 +105,42 @@ def extrapolate_edges(edges, frequency, signal_length):
...
             good_measures.append(d / 2.0)
     return np.mean(good_measures)
 
 
-def get_perfect_frequency2(best_measures, guess_frequency, tollerance=0.15):
-
+def get_perfect_frequency2(
+    best_measures: SpikesArray,
+    guess_frequency: float,
+    tolerance=0.15,
```
BengtRydberg:
```suggestion
    tolerance: float = 0.15,
```

scanomatic/data_processing/norm.py
```diff
@@ -1,5 +1,5 @@
 from enum import Enum
-from typing import Literal, Optional, Sequence, Tuple
+from typing import Literal, Optional, Sequence
```
local-minimum, line 2: iaf sequence from `collections.abc`

scanomatic/data_processing/phenotyper.py
```diff
@@ -5,7 +5,7 @@
 from io import StringIO
 from itertools import chain, product
 from logging import Logger
-from typing import Any, Callable, List, Optional, Union, Tuple
+from typing import Any, Callable, Optional, Union
```
local-minimum, line 8: callable from `collections.abc` iaf

scanomatic/data_processing/strain_handler.py
local-minimum: perhaps enough with `Iterable`.

scanomatic/data_processing/strain_handler.py
skymandr: Rätt sätt verkar vara att importera från collections direkt, inte från abc, tycker jag. Det funkar i alla fall.

local-minimum (approved):
lgtm

skymandr:
For the record, before starting all this we had
```
Found 182 errors in 50 files (checked 162 source files)
```
Now we have
```
Found 171 errors in 38 files (checked 163 source files)
```
...

Not quite working backwards but still...

BengtRydberg (approved):
LGTM

PR #82, opened 2021-11-18
Issue 76 unit tests

### LINKED ISSUE #76: Unit tests should pass

skymandr:
Unit tests run.

**yay!**

Resolves #76 

To clarify: Most of the fixes in the PR should be reasonable, but a few that are needed to actually make the tests pass are probably not. I'll mark the suspicious pieces in the code.

The suspicious lines also high light the underlying errors in the integration tests, which is a bonus.

### DIFF
--- a/scanomatic/generics/abstract_model_factory.py
+++ b/scanomatic/generics/abstract_model_factory.py
@@ -1175,26 +1175,26 @@ class SerializationHelper:
             )
 
     @staticmethod
-    def serialize(obj, dtype):
+    def serialize(obj, dtype) -> Optional[str]:
         if obj is None:
             return None
 
         elif isinstance(dtype, type) and issubclass(dtype, Enum):
-            return obj.name
+            return str(obj.name)
 
         elif dtype is _SectionsLink:
-            return pickle.dumps(obj)
+            return str(pickle.dumps(obj))
 
         elif dtype in (int, float, str, bool):
             return str(obj)
 
         elif isinstance(dtype, types.FunctionType):
-            return pickle.dumps(dtype(serialize=obj))
+            return str(pickle.dumps(dtype(serialize=obj)))
 
         else:
             if not isinstance(obj, dtype):
                 obj = dtype(obj)
-            return pickle.dumps(obj)
+            return str(pickle.dumps(obj))
 
     @staticmethod
     def isvalidtype(o, dtype) -> bool:
--- a/scanomatic/generics/abstract_model_factory.py
+++ b/scanomatic/generics/abstract_model_factory.py
@@ -1030,7 +1030,7 @@ class Serializer:
         if not self._has_section_head(model):
             raise ValueError("Need a section head for serialization")
 
-        with LinkerConfigParser(id=id(model)) as conf:
+        with LinkerConfigParser(id=id(model), allow_no_value=True) as conf:
             conf = self.serialize_into_conf(
                 model,
                 conf,
@@ -1042,12 +1042,6 @@ class Serializer:
             )
 
     def serialize_into_conf(self, model, conf, section):
-        # self._logger.info("Serializing {0} into '{1}' of {2}".format(
-        #     model,
-        #     section,
-        #     conf,
-        # ))
-
         if conf.has_section(section):
             conf.remove_section(section)
 
--- a/scanomatic/generics/abstract_model_factory.py
+++ b/scanomatic/generics/abstract_model_factory.py
@@ -1291,7 +1291,9 @@ class SerializationHelper:
 
         elif isinstance(dtype, types.FunctionType):
             try:
-                return dtype(enforce=pickle.loads(serialized_obj.encode()))
+                return dtype(enforce=pickle.loads(
+                    serialized_obj.encode('iso-8859-1'),
+                ))
             except (pickle.PickleError, EOFError):
                 logging.exception(
                     f"Could not parse {serialized_obj} with type {dtype}",
@@ -1312,7 +1314,7 @@ class SerializationHelper:
 
         else:
             try:
-                return pickle.loads(serialized_obj.encode())
+                return pickle.loads(serialized_obj.encode('iso-8859-1'))
             except (pickle.PickleError, TypeError, EOFError):
                 logging.exception(
                     f"Could not parse {serialized_obj} with type {dtype}",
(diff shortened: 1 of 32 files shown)


scanomatic/generics/abstract_model_factory.py
```diff
@@ -1181,26 +1174,26 @@ def serialize_structure(obj, structure):
...
 
         elif isinstance(dtype, types.FunctionType):
-            return pickle.dumps(dtype(serialize=obj))
+            return pickle.dumps(dtype(serialize=obj)).decode('iso-8859-1')
 
         else:
             if not isinstance(obj, dtype):
                 obj = dtype(obj)
-            return pickle.dumps(obj)
+            return pickle.dumps(obj).decode('iso-8859-1')
```
skymandr, line 1196: I don't think these are reasonable, at least not without proper encode.

scanomatic/generics/abstract_model_factory.py
```diff
@@ -1306,8 +1299,8 @@ def unserialize(serialized_obj: Union[str, bool, Generator], dtype):
 
         else:
             try:
-                return pickle.loads(serialized_obj)
-            except (pickle.PickleError, TypeError):
+                return pickle.loads(serialized_obj.encode())
+            except (pickle.PickleError, TypeError, EOFError):
```
skymandr, line 1303: I think the `encode()` is fine, but maybe not the `EOFError`..?

scanomatic/generics/abstract_model_factory.py
```diff
@@ -1181,26 +1174,26 @@ def serialize_structure(obj, structure):
...
 
     @staticmethod
-    def serialize(obj, dtype):
+    def serialize(obj, dtype) -> Optional[str]:
         if obj is None:
             return None
 
         elif isinstance(dtype, type) and issubclass(dtype, Enum):
-            return obj.name
+            return str(obj.name)
```
skymandr, line 1182: I think this is fine.

scanomatic/generics/abstract_model_factory.py
```diff
@@ -1030,7 +1029,7 @@ def serialize(self, model):
         if not self._has_section_head(model):
             raise ValueError("Need a section head for serialization")
 
-        with LinkerConfigParser(id=id(model)) as conf:
+        with LinkerConfigParser(id=id(model), allow_no_value=True) as conf:
```
skymandr, line 1032: Other instances of `LinkerConfigParser` have this set. It doesn't work without it, so I guess it should be there..?

scanomatic/generics/abstract_model_factory.py
BengtRydberg: kan tycka det känns lite märkligt att metoden tar dtype som input parameter, är det nödvändigt, eller det kanske är så att dtype inte behöver överensstämma med type av obj, vilket rad 1195 verkar antyda? Hade ju varit tydligare om man  kunde typehinta obj till Optional[Union[vad det nu kan vara]]

skymandr: Det är många saker i den här delen av koden som känns som att de kanske är överdesignade, men jag vågar mig nog inte på några större refaktoreringar innan jag tror att det funkar, vilket jag är rätt säker på att det integör, även om unit-testerna går igenom (integrationstesterna visar det, tycker jag).

scanomatic/generics/abstract_model_factory.py
local-minimum: I have some weak memory of having to resort to latin 1, so it might actually not be wrong... were there tests with old pickled things that were actually loaded when decoded like this?

skymandr: I think so, but I think we should take a short session to discuss this in person, so we can be reasonably sure we are on the same page.

scanomatic/generics/abstract_model_factory.py
local-minimum: I don't much remember what that linker config parser does, but some magic with references within configs. I guess we have to chance that allowing no value is acceptable.

scanomatic/generics/abstract_model_factory.py
local-minimum: jag underar seriöst över detta det verkar ju inte vara en instans av en enum utan en typ och då finns ju inte `obj.name`

```python
In [1]: from enum import Enum

In [2]: class C(Enum):
   ...:     D = 1
   ...: 

In [3]: C.name
----------------------------------------------------------------
AttributeError                 Traceback (most recent call last)
<ipython-input-3-bde6c5522d40> in <module>
----> 1 C.name

/usr/lib/python3.9/enum.py in __getattr__(cls, name)
    403             return cls._member_map_[name]
    404         except KeyError:
--> 405             raise AttributeError(name) from None
    406 
    407     def __getitem__(cls, name):

AttributeError: name

In [4]: isinstance(C, type)
Out[4]: True

In [5]: isinstance(C.D, type)
Out[5]: False
```

scanomatic/generics/abstract_model_factory.py
local-minimum: It might be fine, but perhaps logging exception so we at least can track the failings.

scanomatic/generics/abstract_model_factory.py
local-minimum: I think I understand what it does now. What a convoluted thing. Who wrote this 😎 
Maybe replace that if-block with
```python
if isinstance(obj, Enum):
    return obj.name
```
That should be properly typed and all.

scanomatic/generics/abstract_model_factory.py
skymandr: The `EOFError` is an indication that something went wrong earlier in the chain (the symptom is a zero length byte-array I think), but so are the other errors, so it is not unique in that respect. I'll add logging.

scanomatic/generics/abstract_model_factory.py
skymandr: Sure, if you think that's safe.

scanomatic/generics/abstract_model_factory.py
skymandr: All other instances (I think) where it is used this is set to `True`. There might be a reason why not in this case, but... Hopefully that will become clear when we try to fix the integration tests.

local-minimum (commented):
Lgtm, but please add some asserts to the model loading test. eg

```python
assert model.image_data_output_measure is expect_measure
assert model.cell_count_calibration == expect_calibration
assert isinstance(model.grid_model, GridModel)

local-minimum (approved):

PR #87, opened 2021-11-29
add missing type ignore

skymandr:
This single line of code is responsible for going from this
```console
Found 106 errors in 23 files (checked 164 source files)
```
to this
```console
Found 783 errors in 72 files (checked 164 source files)
```
by allowing mypy to not skip things....

While continuing to fix mypy errors is a good thing, this demonstrates that a more structured approach is probably needed.

**Please also not that the build is failing. I haven't looked into why yet!**

### DIFF
--- a/scanomatic/server/rpcjob.py
+++ b/scanomatic/server/rpcjob.py
@@ -5,7 +5,7 @@ from threading import Thread
 from time import sleep
 
 import psutil
-import setproctitle
+import setproctitle  # type: ignore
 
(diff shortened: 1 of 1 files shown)


local-minimum (approved):
lgtm, build fail seemed to be a time out so I restarted it. perhaps it works now 🤷🏾

skymandr:
> lgtm, build fail seemed to be a time out so I restarted it. perhaps it works now 🤷🏾

It did.

PR #89, opened 2021-12-17
Fix various type issues and iterators

local-minimum:
Unfortunately this reverts the fix that created the `analysis` folder because it was not done correctly before. the output folder field may not include `/`

### DIFF
--- a/scanomatic/server/analysis_effector.py
+++ b/scanomatic/server/analysis_effector.py
@@ -1,3 +1,4 @@
+import errno
 import os
 import time
 from typing import Optional
@@ -19,6 +20,7 @@ from scanomatic.models.factories.fixture_factories import (
 )
 from scanomatic.models.factories.scanning_factory import ScanningModelFactory
 from scanomatic.models.rpc_job_models import JOB_TYPE, RPCjobModel
+from scanomatic.models.scanning_model import ScanningModel
 
 from . import proc_effector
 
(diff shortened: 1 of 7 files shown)


scanomatic/server/analysis_effector.py
```diff
@@ -108,6 +110,10 @@ def total(self) -> int:
             return self._first_pass_results.total_number_of_images
         return -1
 
+    @property
+    def ready_to_start(self) -> bool:
```
skymandr, line 114: Du använder inte denna..?

scanomatic/server/compile_effector.py
```diff
@@ -180,9 +180,9 @@ def __next__(self):
             is COMPILE_ACTION.InitiateAndSpawnAnalysis
         ):
             self._spawn_analysis()
-            self.enact_stop()
+            self.enact_stop()  # Raises StopIteration
```
skymandr, line 183: Varför inte bara `raise`?

skymandr (approved):
Looks good, a few minor questions.

skymandr:
> Unfortunately this reverts the fix that created the `analysis` folder because it was not done correctly before. the output folder field may not include `/`

Reopen that issue?

scanomatic/server/analysis_effector.py
local-minimum: in the new test. wanted to have a separate test to show setup is ready

scanomatic/server/compile_effector.py
local-minimum: yes why not... I'll move the raise here

scanomatic/server/analysis_effector.py
skymandr: ah, right. saw the test, but was wondering if it was yagni since you don't use it operationally.

local-minimum:
> > Unfortunately this reverts the fix that created the `analysis` folder because it was not done correctly before. the output folder field may not include `/`
> 
> Reopen that issue?

Not needed, it will have to be solved in the integration tests I believe.

PR #90, opened 2021-12-20
Add more models

local-minimum:
Adds all models that are handled by factories as far as I can see.

Also

- Fixed a spelling typo
- Changed 2 enum creations so that they get typed properly but on purpose keep them as they are
- Added handling of new serializer not asking what should be serialized for special Application model that did funny things

Note

- Existing thing is that some tuples are serialized as lists, though they are inited as tuples. This seems to be as SoM works currently...

### DIFF
--- a/scanomatic/io/jsonizer.py
+++ b/scanomatic/io/jsonizer.py
@@ -1,18 +1,52 @@
 import json
-from enum import Enum, unique
 import logging
-from typing import Any, Type, Union
 from collections.abc import Callable
+from enum import Enum, unique
+from typing import Any, Type, Union
+
 import numpy as np
 
 from scanomatic.generics.model import Model
+from scanomatic.io.power_manager import POWER_MANAGER_TYPE, POWER_MODES
+from scanomatic.models.analysis_model import COMPARTMENTS, MEASURES, VALUES
+from scanomatic.models.compile_project_model import COMPILE_ACTION, FIXTURE
 from scanomatic.models.factories.analysis_factories import (
+    AnalysisFeaturesFactory,
     AnalysisModelFactory,
-    GridModelFactory,
+    GridModelFactory
 )
-from scanomatic.models.analysis_model import COMPARTMENTS, VALUES, MEASURES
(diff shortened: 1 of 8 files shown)


scanomatic/io/jsonizer.py
```diff
@@ -31,9 +65,37 @@ class JSONEncodingError(JSONSerializationError):
 
 
 MODEL_CLASSES: dict[str, Callable[..., Model]] = {
-    "AnalysisModel": AnalysisModelFactory.create,
+    # From analysis_factories.py
```
skymandr, line 68: Thanks for adding comments!

scanomatic/models/factories/settings_factories.py
```diff
@@ -176,7 +176,8 @@ class ApplicationSettingsFactory(AbstractModelFactory):
     @classmethod
     def create(cls, **settings) -> settings_models.ApplicationSettingsModel:
         cls.populate_with_default_submodels(settings)
-
+        if 'versions' in settings:
+            del settings['versions']
```
skymandr, line 180: hmm?

skymandr (approved):
Ser bra ut. En fråga dock.

scanomatic/models/factories/settings_factories.py
local-minimum: The model has a field version but it may not be supplied upon creation and should instead always be constructed from the current default creation inside the model's init function. We might be able to later do away completely with this versions thing that is responsible for some backwards compatibility, but I dare not now.

skymandr: I see. And it is `versions` not `version`? (In your explanation you seem to imply it is a single version per model.)

local-minimum: yes well it could become whatever, but I'd rather keep it and make an issue for investigating removing it.

local-minimum: https://github.com/Scan-o-Matic/scanomatic-standalone/issues/92

PR #91, opened 2021-12-20
types and deprecations

### LINKED ISSUE #42: Validate use of packages
# 📋 Info

Some packages don't have proper type-hinting / VSC doesn't find what is referenced, so need to be manually checked or increased test coverage to ensure they are used properly.

# 🏁 DoD
- [x] `from scipy`
- [x] `import psutil`
- [x] `import setproctitle`
- [x] `from skimage`

skymandr:
Resolves #42 (All four mentioned packages checked). Major fixes includes fixing deprecation of certain `signal` members.

In addition adds a lot of types. I think the types I have put in are mostly correct, but some of these types cause new mypy-conflicts. Fixing those conflicts would make this PR very messy, wherefore I have made no attempt to do so, however, I think having them is better than not having them, since they make finding inconsistencies easier, even if in some cases it is the types and not the code that is wrong. If you feel this has overstepped the bounds of decency, please say so, and I can make a PR with only the minor tweaks needed for #42 instead.


scanomatic/qc/phenotype_results.py
```diff
@@ -1207,26 +1212,26 @@ def _animation():
...
         while index < plate.shape[-1]:
             im.set_data(plate[..., index])
-            ax.set_title("Time {0}".format(index))
+            _ax.set_title("Time {0}".format(index))
             index += 1
 
             yield
 
-    return _animation()
+    return _animation(ax)
```
local-minimum, line 1220: have you tested that this works? I Thought I attempted it but there was some issue

local-minimum: I think theres a readme or similar how to apply it to any function that plots and updates a figure

local-minimum (changes requested):

scanomatic/data_processing/calibration.py
```diff
@@ -106,7 +106,7 @@ def wrapped(identifier, *args, **kwargs):
 
 
 @_validate_ccc_edit_request
-def is_valid_edit_request(identifier):
+def is_valid_edit_request(identifier) -> bool:
```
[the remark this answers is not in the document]
skymandr: Jo, fast det är ju inte direkt en feature att den returnerar `True` jämt väl? Eller finns det nån magi i wrappern som gör att det är så?

scanomatic/qc/phenotype_results.py
skymandr: I changed this because VSC was unhappy with using ax the way it was used. To me it doesn't look suspicious (but on the other hand it didn't really look suspicious before my change either), but I will try it out.

scanomatic/data_processing/calibration.py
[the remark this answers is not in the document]
local-minimum: nej det va bara fyi, tycker det ska stå bool här iaf

scanomatic/data_processing/growth_phenotypes.py
```diff
@@ -315,7 +356,12 @@ def generation_time_error(derivative_errors, index, **kwargs):
     return derivative_errors[index]
 
 
-def generation_time_when(flat_times, index, linregress_extent, **kwargs):
+def generation_time_when(
+    flat_times,
+    index: int,
+    linregress_extent,
```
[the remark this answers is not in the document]
skymandr: Fixat.

scanomatic/qc/phenotype_results.py
skymandr: Reverterade delvis. Se koden.

local-minimum (approved):

Issue #78, opened 2021-11-10
System tests should pass


skymandr:
Update fixtures to use new format

PR #95, opened 2021-12-20
Integrate jsonizer

local-minimum:
# Features

- `jsonizer` extended and integrated
- validation broken out of model factories
- removes old serialization
- removes various unused code bits
- removes the `FIELD_TYPES` from models

# ToDo

- ~Update paths and patterns to give files new names to avoid bad loads?~ No lets keep paths
- `AnalysisModelFactory.set_absolute_path` is really bad since invalidates model.  Will have to deal with this when returning to get integration tests to work. Part of #77
- Only updated unit test fixtures. Letting the integration and system test fixtures remain of old style until we return to working on them. Part of #77 & #78


scanomatic/generics/abstract_model_factory.py
```diff
@@ -491,885 +403,45 @@ def populate_with_default_submodels(cls, obj: Union[dict, Model]) -> None:
...
-                max_val = getattr(max_model, attr)
-
-                if min_val is not None and val < min_val:
-                    setattr(model, attr, min_val)
-                elif max_val is not None and val > max_val:
-                    setattr(model, attr, max_val)
-
-    @classmethod
-    def _correct_type_and_in_bounds(
+    def contains_model_type(
```
local-minimum, line 406: tested as part of the validation tests

scanomatic/image_analysis/first_pass_image.py
```diff
@@ -36,8 +38,35 @@ def get_image_scale(im):
         return invalid_scale
 
 
-def _get_rotated_vector(x, y, rotation):
-    return x * np.cos(rotation), y * np.sin(rotation)
+# def _get_rotated_vector(x, y, rotation):
```
local-minimum, line 41: this was only used in commented out code so will keep until commented out code is removed

scanomatic/image_analysis/first_pass_image.py
```diff
@@ -36,8 +38,35 @@ def get_image_scale(im):
...
+#    return x * np.cos(rotation), y * np.sin(rotation)
+
+
+def get_local_fixture_for_image(identifier, image_identifier):
```
local-minimum, line 45: this just moved here

scanomatic/io/first_pass_results.py
```diff
@@ -84,41 +87,40 @@ def load_scanner_instructions(self, path: Optional[str] = None):
...
         self._reindex_plates(images)
 
+        models = copy(images)
         if sort_mode is FIRST_PASS_SORTING.Time:
-            self._image_models = list(
-                CompileImageAnalysisFactory.copy_iterable_of_model_update_indices(  # noqa: E501
-                    images,
-                )
-            )
+            for (index, m) in enumerate(sorted(
```
local-minimum, line 109: moved the functions here that was previously called

scanomatic/io/jsonizer.py
```diff
@@ -245,3 +246,133 @@ def object_hook(obj: dict) -> Union[dict, Enum, Model]:
 
 def loads(s: Union[str, bytes]) -> Any:
     return json.loads(s, object_hook=object_hook)
+
```
local-minimum, line 249: New implementations of things that previously was located in the abstract factory file.

scanomatic/models/analysis_model.py
```diff
@@ -49,9 +48,33 @@ class VALUES(Enum):
     Cell_Estimates = 2
 
 
+class AnalysisModelFields(Enum):
```
local-minimum, line 51: These explicit enums replaces the previous `FIELD_TYPE`, a bit more wordy but also helped me find some existing typos in validation functions.

scanomatic/models/factories/analysis_factories.py
```diff
@@ -26,57 +24,15 @@ class GridModelFactory(AbstractModelFactory):
     }
 
     @classmethod
-    def _validate_use_utso(cls, model):
```
local-minimum, line 29: These have now moved into new files in the `validators` folder.

scanomatic/models/factories/analysis_factories.py
```diff
@@ -145,149 +100,28 @@ def all_keys_valid(cls, keys):
...
-        if (
-            cls._is_file(model.compilation)
-            and os.path.abspath(model.compilation) == model.compilation
-        ):
-            return True
-        return model.FIELD_TYPES.compilation
-
-    @classmethod
-    def _validate_compilation_instructions_file(
+    def set_default(
```
local-minimum, line 106: this came from the abstract factory class since it was only used here.

scanomatic/models/validators/analysis_model.py
```diff
@@ -0,0 +1,134 @@
...
+    is_pinning_formats,
+    is_real_number,
+    is_safe_path,
+    is_tuple_or_list
+)
+
+ValidationResult = Union[Literal[True], AnalysisModelFields]
+
+
+def validate_compilation_file(model: AnalysisModel) -> ValidationResult:
```
local-minimum, line 18: and here are were the specific validators ended up, they should only have minor changes to them beyond better typing.

scanomatic/models/validators/tools.py
```diff
@@ -0,0 +1,107 @@
+from enum import Enum
```
local-minimum, line 1: most things in here were just moved here, but I wrote tests for them anyhow since it seemed a bit more important than the specific validator functions.

scanomatic/models/validators/validate.py
```diff
@@ -0,0 +1,92 @@
+from collections.abc import Iterator
```
local-minimum, line 1: This is a reimplementation of previous validation logic and should be considered new code

tests/unit/models/factories/data/analysis.model
```diff
@@ -1,77 +1,41 @@
...
-
-[GRIDDING]
-use_utso = True
-median_coefficient = 0.99
-grid
-reference_grid_folder
-manual_threshold = 0.05
-gridding_offsets
-
+{
```
local-minimum, line 1: These are to the best of my understanding exact translations of previous serialized data. but a bit tricky to do so could well be a typo or so.

scanomatic/image_analysis/first_pass_image.py
skymandr: There is an issue for this, right?

scanomatic/models/validators/scanning_aux_info_model.py
```diff
@@ -0,0 +1,66 @@
...
+)
+
+
+ValidationResult = Union[Literal[True], ScanningAuxInfoModelFields]
+
+
+def validate_stress_level(model: ScanningAuxInfoModel) -> ValidationResult:
+    if not isinstance(model.stress_level, int):
+        return ScanningAuxInfoModelFields.stress_level
+    elif model.stress_level == -1 or model.stress_level > 0:
```
skymandr, line 20: why not `is_int_positive_or_neg_one`?

scanomatic/image_analysis/first_pass_image.py
local-minimum: Now there is https://github.com/Scan-o-Matic/scanomatic-standalone/issues/97

scanomatic/models/validators/scanning_aux_info_model.py
local-minimum: dunno (not written now), but will fix

skymandr (changes requested):
That's all I have time for this year. Will finish review after the holidays!

scanomatic/models/validators/tools.py
```diff
@@ -0,0 +1,107 @@
...
+        pass
+
+    return False
+
+
+def is_pinning_formats(pinning_formats: Any) -> bool:
+    try:
+        return all(
+            pinning_format is None
+            or pinning_format is False
```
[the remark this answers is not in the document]
local-minimum: I don't know, but probably a dangerous idea to change it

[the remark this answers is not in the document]
skymandr: Probably... I guess we have used `if pinning_format:` or similar in the past...

scanomatic/io/jsonizer.py
```diff
@@ -245,3 +246,133 @@ def object_hook(obj: dict) -> Union[dict, Enum, Model]:
...
+
+T = TypeVar('T')
+
+
+def copy(o: T) -> T:
```
skymandr, line 254: I don't quite understand this typing. Could you elaborate?

scanomatic/io/jsonizer.py
```diff
@@ -245,3 +246,133 @@ def object_hook(obj: dict) -> Union[dict, Enum, Model]:
...
+
+def dump(
+    model: Model,
+    path: Union[str, Path],
+    overwrite: bool = False,
+) -> bool:
+    if overwrite:
+        model = merge_into(load(path), model)
+    try:
+        with open(path, 'w') as fh:
```
skymandr, line 309: I wonder what it says about the provenance of a programmer, whether they prefer `fp` or `fh`? I have some very unsubstantiated ideas. (I think both are fine, just curious.)

scanomatic/models/validators/compile_instructions_model.py
```diff
@@ -0,0 +1,65 @@
...
+        return True
+    return CompileInstructionsModelFields.path
+
+
+def validate_fixture(model: CompileInstructionsModel) -> ValidationResult:
+    if model.fixture_type is FIXTURE.Local:
+        if os.path.isfile(
+            os.path.join(
+                model.path,
+                Paths().experiment_local_fixturename,
```
skymandr, line 45: A shame to have to instantiate the `Paths` on each check, but I guess it doesn't happen too often.

scanomatic/models/validators/compile_instructions_model.py
```diff
@@ -0,0 +1,65 @@
...
+            ),
+        ):
+            return True
+        else:
+            return CompileInstructionsModelFields.fixture_type
+    elif model.fixture_type is FIXTURE.Global:
+        if model.fixture_name in Fixtures():
```
skymandr, line 52: Same here with `Fixtures`.

tests/unit/models/validators/test_tools.py
```diff
@@ -0,0 +1,195 @@
...
+    assert is_coordinates(obj) is expected
+
+
+@pytest.mark.parametrize("obj,expected", [
+    (-1, True),
+    (0, True),
+    (2.42, True),
+    (1j, False),
+    ('a', False),
+    (None, False),
```
skymandr, line 104: perhaps also test some non-finite?

tests/unit/models/validators/test_validate.py
```diff
@@ -0,0 +1,147 @@
...
+        FixtureFactory.create(
+            grayscale=GrayScaleAreaModelFactory.create(),
+        ),
+        {
+            "plates": (FixturePlateFactory.create(), None),
+        },
+        True,
+    ),
+))
+def test_validate(model: Model, updates: dict[str, Any], expect: bool):
```
skymandr, line 91: Maybe a few of these cases could do with a comment?

skymandr: (Here and elsewhere, but I'm not sure if it would help much.)

scanomatic/models/validators/validate.py
```diff
@@ -0,0 +1,92 @@
...
+
+class UnknownFields(Enum):
+    UNKNOWN_FIELD = auto()
+
+
+def _produce_validation_error(fields: Optional[Type[Enum]], field: str) -> Enum:
+    if fields is None:
+        return UnknownFields.UNKNOWN_FIELD
+    if field == "name":
+        return fields["_name"]
```
skymandr, line 22: What is the logic here?

skymandr (commented):
A few more comments, but mostly it looks good. I have one more thing to check though.

scanomatic/io/jsonizer.py
local-minimum: whatever type you put in you get the same one out.

scanomatic/io/jsonizer.py
local-minimum: no clue

scanomatic/models/validators/compile_instructions_model.py
local-minimum: yeah it really shouldn't be a class I think, but a module

scanomatic/models/validators/validate.py
local-minimum: enums already have `name` attribute so it's really problematic that there were enum values called `name` too... this works around that.

tests/unit/models/factories/test_analysis_factory.py
```diff
@@ -34,15 +81,11 @@ def test_model_has_ccc_id(self, analysis_model):
...
+        serial = dumps(analysis_model)
+        assert len(json.loads(serial)) == 2
 
     def test_model_can_deserialize(self, analysis_serialized_object):
-        result = AnalysisModelFactory.get_serializer().load_serialized_object(
-            analysis_serialized_object,
-        )
-        assert len(result) == 1
-        model = result[0]
+        model: AnalysisModel = loads(analysis_serialized_object)
```
skymandr, line 88: As we've discussed previously, this test seemed to actually only test things that were set to defaults (row 90+). If this is still the case, then adding something like email (and perhaps not check the defaults at all?) would be a welcome addition, while we remember it. This may apply to more tests (though I cannot think of any), but I know it was the case for this one in particular.

scanomatic/models/validators/compile_instructions_model.py
local-minimum: agree, but not now. I think it may be using a singleton type of initiation though so it shouldn't be too expensive.

skymandr (changes requested):
One final thing that would be nice to have done, lest we forget about it.

tests/unit/models/validators/test_validate.py
local-minimum: added here at least

tests/unit/models/factories/test_analysis_factory.py
local-minimum: fixed

skymandr (approved):
Ser bra ut!

PR #102, opened 2022-01-12
Fix scan writing

local-minimum:
With this PR only analysis test remain to be repaired.

Bonus fixes so logging is better in attempt to find out where things go wrong.


scanomatic/io/rpc_client.py
```diff
@@ -179,12 +182,12 @@ def local(self) -> bool:
...
-    def userID(self):
-        return self._userID
+    def user_id(self):
+        return self._user_id
 
-    @userID.setter
-    def userID(self, value):
-        self._userID = value
+    @user_id.setter
+    def user_id(self, value):
```
skymandr, line 189: maybe add `-> None` for clarity?

scanomatic/ui_server/data_api.py
```diff
@@ -17,16 +16,16 @@
 from scanomatic.data_processing.norm import NormState
 from scanomatic.data_processing.project import path_has_saved_project_state
 from scanomatic.image_analysis.first_pass_image import FixtureImage
-from scanomatic.image_analysis.grayscale import (
+from scanomatic.image_analysis.grayscale import \
```
skymandr, line 19: Oh? Interesting.

scanomatic/io/scanstore.py
```diff
@@ -22,7 +22,7 @@ def add_scan(self, project_fullname, scan):
             scan.timedelta.total_seconds(),
         )
         path = os.path.join(self._path, project_fullname, filename)
-        with open(path, 'w') as f:
+        with open(path, 'wb') as f:
```
skymandr, line 25: Interesting! I wasn't aware this was different in Python 3, but I guess it makes sense with the new string/bytes division.

skymandr (approved):
Looks good, though see comments.

scanomatic/ui_server/data_api.py
local-minimum: That's the automatic formatting script does that unfortunately but will fix

scanomatic/io/scanstore.py
local-minimum: the difference is that what we attempt to write is now bytes so we need to write it as bytes too.

scanomatic/io/rpc_client.py
```diff
@@ -37,28 +37,32 @@ def sanitize_communication(obj):
         return obj
 
 
-def get_client(host=None, port=None, admin=False, log_level=None):
-    appCfg = app_config.Config()
+def get_client(
+    host=None,
+    port=None,
+    admin=False,
+    log_level=None,
```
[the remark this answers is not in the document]
local-minimum: better remove all the arguments instead

scanomatic/io/rpc_client.py
local-minimum: removed them instead

scanomatic/io/scanstore.py
skymandr: Yes, I hadn't registered that this was the case in 3, but it makes sense.

skymandr: Yes, I hadn't registered that this was the case in 3, but it makes sense.

scanomatic/io/rpc_client.py
```diff
@@ -77,25 +71,24 @@ def __init__(self, host, port, userID=None, log_level=None):
...
+        else:
+            self._logger.warning(
+                "Can't launch because server is {0}".format(
+                    ['not local', 'online'][self.online],
                 )
+            )
 
-        elif key in self._allowedMethods() and self._client is not None:
+    def __getattr__(self, key):
+        if key in self._allowedMethods() and self._client is not None:
```
skymandr, line 86: Since you did `s/userID/user_id/` you may want to consider doing the same to `allowedMethods` and userIDdecorator` too.

local-minimum: I'll do that in the next PR that is almost ready.

local-minimum:
> I didn't quite get why `admin` now has the type `str` rather than `bool`, but all in all it looks good (better, even). I commented on a few small things you may want to look at while you're at it.

It might be that it can be both by some obscure magic but I noticed that contents from a file was read and placed on the  `admin` field or a new admin thing was generated and that was a `str`... could be that it can be a bool too 🤷🏾

scanomatic/io/rpc_client.py
skymandr: Tanoshimi.

Issue #51, opened 2021-11-02
Fix MyPy errors


local-minimum:
# 📋 Info

MyPy errors indicate something is strange.

# 🏁 DoD

- [ ] No MyPy error

skymandr:
Things such as scipy/numpy/matplotlib still don't have types, right? so just # type: ignore on these imports..?

local-minimum:
numpy has types in `numpy.typing`. I've introduced some of it.

PR #103, opened 2022-01-12
Fix mypy errors - part 1

e-larsson:
Part of #51

### DIFF
--- a/scanomatic/image_analysis/signal.py
+++ b/scanomatic/image_analysis/signal.py
@@ -637,11 +637,11 @@ def get_closest_signal_pair(
             else:
                 signals.append((p, p2))
 
-    S = np.array(signals)
-    if S.size == 0:
+    signals_arr = np.array(signals)
+    if signals_arr.size == 0:
         raise SignalError("No signal pairs found")
 
-    return S[:, 0], S[:, 1]
+    return signals_arr[:, 0], signals_arr[:, 1]
 
 
 def get_signal_spikes(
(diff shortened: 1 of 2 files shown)


scanomatic/image_analysis/signal.py
```diff
@@ -637,7 +639,7 @@ def get_closest_signal_pair(
 
     S = np.array(signals)
     if S.size == 0:
-        return None, None
+        raise SignalError("No signal pairs found")
```
e-larsson, line 642: The None tuple wasn't handled anywhere

scanomatic/image_analysis/signal.py
```diff
@@ -727,7 +729,7 @@ def get_offset_quality(
     return q.sum()
 
 
-def _get_wave_length_and_errors(s, expected_spikes: int) -> float:
```
e-larsson, line 730: Unused parameter

skymandr: I think these share an interface, hence why it has that argument, but @local-minimum can hopefully elucidate.

local-minimum: I did a lot of yagni back then and probably used it while developing the algorithm to figure something out, but seems reasonable to remove now

scanomatic/image_analysis/signal.py
```diff
@@ -661,7 +663,7 @@ def get_signal_spikes(
     ).astype(int)
 
     s1, s2 = get_closest_signal_pair(d_up, d_down, s1_value=-1, s2_value=1)
-    return (s1 + s2) / 2.0  # (s1 + s2) / 2.0
```
local-minimum, line 664: what a great comment 😆

scanomatic/image_analysis/signal.py
local-minimum: sounds good consider getting rid of the `S` variable as something like `signals_arr` or something too.

local-minimum (approved):

PR #104, opened 2022-01-12
Fix mypy errors - part 2

e-larsson:
Part of #51

### DIFF
--- a/scanomatic/image_analysis/image_basics.py
+++ b/scanomatic/image_analysis/image_basics.py
@@ -7,6 +7,7 @@ from PIL import Image
 from scipy.ndimage import zoom  # type: ignore
 
 from scanomatic.models.analysis_model import IMAGE_ROTATIONS
+from .exceptions import LoadImageError
 
 _logger = Logger("Basic Image Utils")
 
@@ -60,10 +61,11 @@ def Quick_Scale_To_im(
 ) -> np.ndarray:
     if im is None:
         try:
-            im = load_image_to_numpy(path, dtype=np.uint8)
+            im = load_image_to_numpy(str(path), dtype=np.uint8)
         except Exception:
-            _logger.error("Could not open source")
-            return -1
+            msg = f"Could not open source with path {path}"
+            _logger.error(msg)
+            raise LoadImageError(msg)
 
     if scale is None:
         scale = target_dpi / source_dpi
--- a/scanomatic/image_analysis/exceptions.py
+++ b/scanomatic/image_analysis/exceptions.py
@@ -1,0 +1,2 @@
+class LoadImageError(Exception):
+    pass


scanomatic/image_analysis/image_basics.py
```diff
@@ -60,10 +61,11 @@ def Quick_Scale_To_im(
 ) -> np.ndarray:
     if im is None:
         try:
-            im = load_image_to_numpy(path, dtype=np.uint8)
+            im = load_image_to_numpy(str(path), dtype=np.uint8)
         except Exception:
-            _logger.error("Could not open source")
-            return -1
```
e-larsson, line 66: This -1 didn't seem to be handled anywhere

scanomatic/image_analysis/image_basics.py
```diff
@@ -60,10 +61,11 @@ def Quick_Scale_To_im(
...
+            msg = f"Could not open source with path {path}"
+            _logger.error(msg)
```
skymandr, line 67:
```suggestion
            _logger.exception(msg)
```

skymandr (approved):
LGTM, but I think @local-minimum is interested in having a look.

scanomatic/image_analysis/image_basics.py
local-minimum: Agree

scanomatic/image_analysis/image_basics.py
```diff
@@ -60,10 +61,11 @@ def Quick_Scale_To_im(
 ) -> np.ndarray:
     if im is None:
         try:
-            im = load_image_to_numpy(path, dtype=np.uint8)
+            im = load_image_to_numpy(str(path), dtype=np.uint8)
```
local-minimum, line 64: I think it better to before the try raise an `LoadImageError` if `path is None and im is None`

local-minimum (changes requested):

scanomatic/image_analysis/image_basics.py
e-larsson: Fixed

scanomatic/image_analysis/image_basics.py
```diff
@@ -59,11 +60,17 @@ def Quick_Scale_To_im(
     scale: Optional[float] = None,
 ) -> np.ndarray:
     if im is None:
+        if path is None:
+            msg = "No image or path supplied"
+            _logger.exception(msg)
```
local-minimum, line 65:
```suggestion
            _logger.error(msg)
```

scanomatic/image_analysis/image_basics.py
```diff
@@ -59,11 +60,17 @@ def Quick_Scale_To_im(
...
+            raise LoadImageError(msg)
+
         try:
             im = load_image_to_numpy(path, dtype=np.uint8)
         except Exception:
-            _logger.error("Could not open source")
-            return -1
+            msg = f"Could not open source with path {path}"
+            _logger.exception(msg)
```
local-minimum, line 72:
```suggestion
            _logger.error(msg)
```
I don't think we need the traceback in the log here

local-minimum (approved):
lgtm but change the logging level as suggested.

PR #105, opened 2022-01-12
Fix system test pt3

local-minimum:
All system tests pass locally! 🥳


local-minimum:
> I have no objections beyond the regrettable fact that the tests fail on github, and that types for `pandas` are not recognized.

Neither has anything to do with this PR though

skymandr:
> > I have no objections beyond the regrettable fact that the tests fail on github, and that types for `pandas` are not recognized.
> 
> Neither has anything to do with this PR though

In that case this PR was not very well described.

PR #106, opened 2022-01-12
System tests action on GitHub

### LINKED ISSUE #78: System tests should pass

local-minimum:
Resolves #78

### DIFF
--- a/tests/system/test_analysis.py
+++ b/tests/system/test_analysis.py
@@ -76,7 +76,7 @@ def assert_has_job(scanomatic, job_settings):
         )
     )
     tries = 0
-    while tries < 25:
+    while tries < 50:
         payload = requests.get(uri).json()
         if payload.get('instructions'):
             assert (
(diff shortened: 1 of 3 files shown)


tests/system/test_analysis.py
```diff
@@ -76,7 +76,7 @@ def assert_has_job(scanomatic, job_settings):
         )
     )
     tries = 0
-    while tries < 25:
+    while tries < 50:
```
skymandr, line 79: ouch...

skymandr (approved):
LGTM. Do you know how long it takes, approximately?

local-minimum:
> LGTM. Do you know how long it takes, approximately?

image

tests/system/test_analysis.py
local-minimum: it's really the first test that failed because surpassing 25, might have been done on 26th but I figured doubling is a good practice.

PR #108, opened 2022-01-12
Mypy fixes

BengtRydberg:
for a few files, primarily in the data_processing directory. This PR seems to resolve all mypy errors of modules in this directory.

### DIFF
--- a/scanomatic/data_processing/data_bridge.py
+++ b/scanomatic/data_processing/data_bridge.py
@@ -1,3 +1,4 @@
+from typing import List
 import numpy as np
 
 from . import mock_numpy_interface
@@ -36,7 +37,9 @@ class Data_Bridge(mock_numpy_interface.NumpyArrayInterface):
             self.update_source = self._update_to_array
 
         elif isinstance(self._source, dict):
-            plates = [[]] * len(self._source)  # Creates plates and 1D pos
+            plates: List[
+                List[List[List[float]]]
+            ] = [[]] * len(self._source)  # Creates plates and 1D pos
             for p in list(self._source.values()):
 
                 for d1 in p:
--- a/scanomatic/data_processing/data_bridge.py
+++ b/scanomatic/data_processing/data_bridge.py
@@ -1,4 +1,3 @@
-from typing import List
 import numpy as np
 
 from . import mock_numpy_interface
@@ -37,8 +36,8 @@ class Data_Bridge(mock_numpy_interface.NumpyArrayInterface):
             self.update_source = self._update_to_array
 
         elif isinstance(self._source, dict):
-            plates: List[
-                List[List[List[float]]]
+            plates: list[
+                list[list[list[float]]]
             ] = [[]] * len(self._source)  # Creates plates and 1D pos
             for p in list(self._source.values()):
 
--- a/scanomatic/data_processing/norm.py
+++ b/scanomatic/data_processing/norm.py
@@ -90,7 +90,10 @@ def infer_offset(arr):
 
 def get_downsampled_plates(
     data: np.ndarray,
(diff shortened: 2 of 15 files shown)


scanomatic/data_processing/norm.py
```diff
@@ -89,8 +89,8 @@ def infer_offset(arr):
 
 
 def get_downsampled_plates(
-    data,
-    subsampling: Literal['TL', 'TR', 'BL', 'BR'] = "BR",
+    data: np.ndarray,
+    subsampling: Union[List[np.ndarray], Literal['TL', 'TR', 'BL', 'BR']] = "BR",
```
local-minimum: in python 3.9 we don't need `List`, `Tuple`, `Dict` anymore, we can use `list`, `tuple` `dict` directly. Various places below

local-minimum (changes requested):

scanomatic/generics/maths.py
```diff
@@ -14,23 +14,25 @@ def iqr_mean(data: np.ndarray, *args, **kwargs) -> Optional[np.ndarray]:
...
-        data = np.ma.masked_invalid(data)
-
-    data = data[data.mask == False]  # noqa: E712
+    data = get_finite_data(data)
     threshold = int(np.floor(data.size * 0.25))
     data.sort()
     return data[threshold], data[-threshold]
+
+
+def get_finite_data(
```
skymandr, line 31: Maybe add a test?

scanomatic/data_processing/norm.py
BengtRydberg: ok, fixed.

scanomatic/generics/maths.py
BengtRydberg: added unit test for this function.

scanomatic/data_processing/data_bridge.py
```diff
@@ -36,7 +37,9 @@ def _create_array_representation(self, **kwargs):
             self.update_source = self._update_to_array
 
         elif isinstance(self._source, dict):
-            plates = [[]] * len(self._source)  # Creates plates and 1D pos
+            plates: List[
+                List[List[List[float]]]
+            ] = [[]] * len(self._source)  # Creates plates and 1D pos
```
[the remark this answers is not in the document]
BengtRydberg: not used elsewhere in the module so I do not think it is of much help to declare it.

scanomatic/data_processing/phases/segmentation.py
```diff
@@ -1,6 +1,6 @@
 import operator
 from enum import Enum
-from typing import Optional, Union
+from typing import Iterator, Optional, Union, cast
```
local-minimum, line 3: Iterator, Sequence etc are deprecated in `typing` and should be imported from `collections.abc`. It's mostly `Optional`, `TypeVar`, `Union`, `Any` and `cast` that should come from `typing`.

scanomatic/data_processing/subplates.py
```diff
@@ -1,8 +1,12 @@
+from typing import Optional, Union
 import numpy as np
 
 from . import mock_numpy_interface
 
 
+TUPLE_TYPE = Union[tuple[int, int], tuple[int, int, int]]
```
local-minimum, line 7: maybe `PlateShape` or `Plate` is better type name?

BengtRydberg: agree. changed to PlateShape.

scanomatic/data_processing/phases/segmentation.py
BengtRydberg: thanks for the info. fixed.

local-minimum (approved):

PR #109, opened 2022-01-13
Fix mypy errors - part 3

e-larsson:
Part of #51


local-minimum (changes requested):
`analysis_image.py`, `calibration_api.py`, `data_api.py` uses `get_grayscale`. There are probably some more.

local-minimum:
Also note that the `get_grayscales` object is put to `jsonify` in `data_api.py` so this should be checked that it works properly.

local-minimum (approved):

Issue #39, opened 2021-10-29
Various streams are now bytes


local-minimum:
# 📋 Info

They used to be strings, and things will probably not work

# 🏁 DoD

- [ ] check all io... 😓

local-minimum:
Example:
```
scanomatic_1  | Traceback (most recent call last):
scanomatic_1  |   File "/usr/local/lib/python3.9/threading.py", line 973, in _bootstrap_inner
scanomatic_1  |     self.run()
scanomatic_1  |   File "/usr/local/lib/python3.9/threading.py", line 910, in run
scanomatic_1  |     self._target(*self._args, **self._kwargs)
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/server/server.py", line 134, in _run
scanomatic_1  |     self._jobs.sync()
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/server/jobs.py", line 141, in sync
scanomatic_1  |     self.handle_scanners()
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/server/jobs.py", line 150, in handle_scanners
scanomatic_1  |     self._scanner_manager.update()
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/io/scanner_manager.py", line 385, in update
scanomatic_1  |     alive_scanners = get_alive_scanners()
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/io/sane.py", line 25, in get_alive_scanners
scanomatic_1  |     return re.findall(
scanomatic_1  |   File "/usr/local/lib/python3.9/re.py", line 241, in findall
scanomatic_1  |     return _compile(pattern, flags).findall(string)
scanomatic_1  | TypeError: cannot use a string pattern on a bytes-like object
```

skymandr:
Some, but not all, of this is done.

PR #110, opened 2022-01-13
Update opens

local-minimum:
Part of #39 

Manually tried out how commands seem to work now as it is quite hard to actually test most of these changes at the moment.

### DIFF
--- a/scanomatic/data_processing/pheno/save.py
+++ b/scanomatic/data_processing/pheno/save.py
@@ -1,7 +1,7 @@
 import os
 import pickle
 import zipfile
-from io import StringIO
+from io import BytesIO
 from typing import Optional
 
 import numpy as np
@@ -101,7 +101,7 @@ def save_state(
         or not os.path.isfile(p)
         or _do_ask_overwrite(p)
     ):
-        with open(p, 'w') as fh:
+        with open(p, 'wb') as fh:
             pickle.dump(state.phenotype_filter_undo, fh)
 
     p = os.path.join(dir_path, _paths.phenotype_times)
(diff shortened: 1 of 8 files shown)


e-larsson (approved):
LGTM

scanomatic/data_processing/pheno/save.py
```diff
@@ -236,7 +236,7 @@ def zipit(save_functions, data, zip_paths):
     zip_paths.append(
         os.path.join(dir_path, _paths.phenotypes_filter_undo),
     )
-    save_functions.append(lambda x, y: pickle.dump(y, x))
+    save_functions.append(lambda fh, obj: pickle.dump(obj, fh))
```
skymandr, line 239: Thanks!

scanomatic/io/pickler.py
```diff
@@ -45,7 +45,7 @@ def __init__(self):
         self._next = None
         self.io = None
 
-    def __call__(self, line: str):
+    def __call__(self, line: bytes):
```
skymandr, line 48: This file makes me nervous by proxy! Is it still useful, even after moving to json configs?

skymandr (approved):
lgtm

scanomatic/io/pickler.py
local-minimum: unfortunately this is used for saving things that are not `Model`s... so yes this is a very hackish thing that should be removed some day hopefully.

skymandr: Good thing you have a decent store of bubbly beverages for all these upcoming moments of celebration.

PR #111, opened 2022-01-13
Fix mypy errors - part 4

e-larsson:
Part of #51


scanomatic/image_analysis/image_grayscale.py
```diff
@@ -422,8 +432,9 @@ def detect_grayscale(im_trimmed, grayscale: Grayscale):
                     np.save(os.path.join(
                         Paths().log,
                         f"gs_segment_{i}.npy",
+                    ),
```
skymandr, line 435: Strange indentation, better to break before `os.path` and indent accordingly?

BengtRydberg, line 435: den här radbrytningen ser lite knepig ut.

scanomatic/image_analysis/image_grayscale.py
```diff
@@ -437,9 +448,10 @@ def detect_grayscale(im_trimmed, grayscale: Grayscale):
             ),
         )
 
-    if len(grayscale_segment_centers) == 0:
+    if grayscale_segment_centers is None:
```
local-minimum, line 451: they can't have length 0 still?

tools/signal.py
```diff
@@ -0,0 +1,447 @@
...
+
+        for i, s in enumerate([s - int(s) for s in shifts]):
+            if s != 0:
+                signals[i][:] = np.array(
+                    [sign + s * frequencies[i] for sign in signals[i]]
+                )
+        return signals
+
+
+def get_continious_slopes(
```
skymandr, line 69:
```suggestion
def get_continuous_slopes(
```

local-minimum (commented):
generally looks good, only one question

tools/signal.py
```diff
@@ -0,0 +1,447 @@
...
+    new_list = []
+    for i in ideal_signal:
+        delta_i = np.abs(X - i)
+        delta_reciprocal = np.abs(Y - X[delta_i.argmin()])
+        if delta_i.min() == delta_reciprocal.min():
+            new_list += ([X[delta_i.argmin()],
+                Y[delta_reciprocal.argmin()]])
+
+    return np.array(new_list)
+"""
```
skymandr, line 178: Remove and rename `_get_alt_closest` to `_get_closest`?

tools/signal.py
```diff
@@ -0,0 +1,447 @@
...
+    ].ravel()
+    return Z
+
+
+def _get_orphans(X: np.ndarray, shortX: np.ndarray) -> np.ndarray:
+    return X[np.abs(np.subtract.outer(X, shortX)).min(axis=1).astype(bool)]
+
+
+def get_offset_quality(
+    s,
```
skymandr, line 197: This and `w1` aren't very nice. Maybe @local-minimum has ideas for better names? (Here and below.)

skymandr (commented):
Not much to add.

tools/signal.py
local-minimum: nope and the point of `/tools` is to not really maintain the code until we notice we or someone needs it so I would argue to put minimal time fixing things here... perhaps typos but not more.

skymandr: Is the plan in the long run, then, to remove this directory?

tools/signal.py
e-larsson: Fixed

scanomatic/image_analysis/image_grayscale.py
e-larsson: Fixed

scanomatic/image_analysis/image_grayscale.py
e-larsson: I guess that could be possible still. Added it again

scanomatic/image_analysis/image_grayscale.py
e-larsson: Fixat

tools/signal.py
local-minimum: move it to a new repo

local-minimum (approved):

PR #113, opened 2022-01-13
Fix stringio

### LINKED ISSUE #39: Various streams are now bytes

local-minimum:
Resolves #39 

There might be errors with str vs bytes still but they'll have to be dealt with when doing the validation tasks.


skymandr (approved):
Looks good!

PR #115, opened 2022-01-13
Fix models

### LINKED ISSUE #96: `GrayScaleAreaModel` should not define field `values`
# 🐛 Description

The field is part of the base class `Model`:s interface and should not be overridden.

# ✌🏽 Expected behaviour

Should store the values on some other field

# DoD

- [x] Stores values on something else maybe `section_values`
- [x] Integrates all usage
- [x] Update factory so that it handles renaming old `values` to new name (exists helper function)
- [x] Add `values` to reserved words in `models.py`

local-minimum:
Resolves #96 

- Changes name of `values` to `section_values` with backwards compatibility
- Requires sub-models in most cases (except RPCjobModel since it's not deterministic what the content model type should be)

@e-larsson and @BengtRydberg hope this doesn't interfer too much with your stuff.


scanomatic/image_analysis/image_grayscale.py
```diff
@@ -161,8 +162,8 @@ def get_para_trimmed_slice(
 
 
 def get_grayscale(
-    fixture,
-    grayscale_area_model,
+    fixture,  # FixtureImage (but circular import)
```
skymandr, line 165: Could `get_grayscale` move to a different/own file? Worth an issue?

scanomatic/models/scanning_model.py
```diff
@@ -295,7 +296,7 @@ def __init__(
         informed_close_to_end: bool = False,
         compilation_state: COMPILE_STATE = COMPILE_STATE.NotInitialized,
     ):
-
+        assert compile_project_model is not None
```
skymandr, line 299: Same as above, I guess.

scanomatic/models/settings_models.py
```diff
@@ -195,16 +195,22 @@ def __init__(
         scanner_names: Optional[Sequence[str]] = None,
         scanner_sockets: Optional[dict[str, int]] = None,
     ):
+        assert power_manager is not None
```
skymandr, line 198: Same as above.

scanomatic/qc/compilation_results.py
```diff
@@ -1,6 +1,7 @@
 import glob
 import os
 import re
+from typing import Sequence
```
skymandr, line 4: shouldn't this be `collections.abc`?

scanomatic/models/settings_models.py
BengtRydberg, line 198: To me it looks a bit odd to type the class to take optional attributes as input and then directly raise if they are None. Why not just skip the optional typing?

scanomatic/ui_server/data_api.py
```diff
@@ -16,9 +16,8 @@
 from scanomatic.data_processing.norm import NormState
 from scanomatic.data_processing.project import path_has_saved_project_state
 from scanomatic.image_analysis.first_pass_image import FixtureImage
-from scanomatic.image_analysis.grayscale import (
+from scanomatic.image_analysis.grayscale import \
```
skymandr, line 19: I guess we'll have to accept this, and the trailing `,` unless we can get the plugins to work the way we want. Unreasonable to fix these things by hand if the editor then comes to reset it! ;)

skymandr (changes requested):
I'm not really expecting you to change things (beyond the odd small comment that may point to something that is actually wrong), but I would like your comments before proceeding.

scanomatic/image_analysis/image_grayscale.py
local-minimum: It should, or something else should move... I'll make an issue

local-minimum: there's also another function in another file that is called `get_grayscale` and it's a bit confusing.

local-minimum: https://github.com/Scan-o-Matic/scanomatic-standalone/issues/118

scanomatic/models/scanning_model.py
```diff
@@ -130,14 +130,15 @@ def __init__(
...
-        auxillary_info: ScanningAuxInfoModel = ScanningAuxInfoModel(),
+        auxillary_info: Optional[ScanningAuxInfoModel] = None,
         plate_descriptions: tuple["PlateDescription", ...] = tuple(),
         version: str = scanomatic.__version__,
         scanning_program: str = "",
         scanning_program_version: str = "",
         scanning_program_params: Sequence[str] = tuple(),
         cell_count_calibration_id=None,
     ):
+        assert auxillary_info is not None
```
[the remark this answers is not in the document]
local-minimum: There's a reason. There's no safe default immutable default value. Since the models are mutable they really shouldn't be used as default values.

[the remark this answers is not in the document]
local-minimum: I don't think an extra message adds enough, the traceback it would cause is more than enough to know that somehow `auxillariy_info` was `None`.

scanomatic/qc/compilation_results.py
local-minimum: yes must have automatically imported from typing.

scanomatic/ui_server/data_api.py
local-minimum: I can't find it, but feel free to come up with a solution for the plugin

scanomatic/image_analysis/image_grayscale.py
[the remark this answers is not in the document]
local-minimum: I made an issue. Previous circular import refactoring I've done have been .... annoying and exploded in size so rather do that separately.

scanomatic/models/settings_models.py
local-minimum: See my answer to @skymandr  above

BengtRydberg (approved):

scanomatic/models/scanning_model.py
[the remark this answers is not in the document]
skymandr: But not really that it was unexpected, looking at the function signature.

scanomatic/ui_server/data_api.py
skymandr: What plugin are you using?

scanomatic/models/scanning_model.py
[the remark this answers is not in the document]
local-minimum: It must be keywords only arguments by design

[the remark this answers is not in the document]
skymandr: Ouch... But I guess it is safe to make this kind of assert wherever mypy complains, at least.

skymandr (approved):

scanomatic/models/scanning_model.py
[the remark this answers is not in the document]
local-minimum: It's safe to make exactly these I made here at least. Wouldn't dare generalize that much.

PR #116, opened 2022-01-13
Fix mypy errors - part 5

e-larsson:
Part of #51 

The intent was to fix mypy errors in `image_analysis/image_fixture.py` but I think most of the errors originate from the `FixtureImage` init function being too generous in how one can instantiate a `FixtureImage` object.

`FixtureImage` is only used in one place with one of the classmethods.

There are more things which could be improved, such as most of the functions could be moved out of the class, but I spent enough time on this...


scanomatic/image_analysis/image_fixture.py
```diff
@@ -9,112 +8,103 @@
...
+        try:
+            image = load_image_to_numpy(image_path, dtype=np.uint8)
 
-        if (
-            os.path.isfile(pattern_image_path) is False
-            and resource_paths is not None
-        ):
-            pattern_image_path = os.path.join(
-                resource_paths.images,
-                os.path.basename(pattern_image_path),
```
e-larsson, line 47: I didn't understand what this was for and it doesn't seem to be used so I just scrapped it. I'll add it back in if anyone objects.

scanomatic/image_analysis/first_pass_image.py
```diff
@@ -259,27 +260,24 @@ def run_marker_analysis(self, markings=None):
...
-            y_positions_correct_scale
-         ) = im_analysis.find_pattern(markings=markings)
+        try:
+            (
+                x_positions_correct_scale,
+                y_positions_correct_scale
+            ) = im_analysis.find_pattern(markings=markings)
+
+        except FixtureImageError:
+            _logger.error("No markers found")
```
local-minimum, line 276: should it really except this and continue? missing a return? or perhaps it is better that it crashes for now since without finding the pattern nothing really can be done.

scanomatic/image_analysis/image_fixture.py
```diff
@@ -9,112 +8,103 @@
...
-                ),
+    @classmethod
+    def from_image(
+        cls,
+        image: np.ndarray,
+        pattern_image_path: str,
+        scale: float = 1.0,
+    ) -> "FixtureImage":
+        if len(image.shape) > 2:
+            image = image[:, :, 0]
```
local-minimum, line 37:
```suggestion
            # Use first channel of color image
            image = image[:, :, 0]
```

scanomatic/image_analysis/image_fixture.py
```diff
@@ -9,112 +8,103 @@
...
-        )
+                dtype=np.uint8,
+            )
+        except IOError:
+            msg = f"Could not open orientation guide image at {pattern_image_path}"  # noqa: E501
+            _logger.error(msg)
+            raise FixtureImageError(msg)
+
+        if len(pattern_image.shape) > 2:
+            pattern_image = pattern_image[:, :, 0]
```
local-minimum, line 50:
```suggestion
            # Use first channel of color image
            pattern_image = pattern_image[:, :, 0]
```

scanomatic/image_analysis/image_fixture.py
```diff
@@ -9,112 +8,103 @@
...
-        conversion_factor: float = 1.0,
+        conversion_factor: float
     ) -> None:
-
+        _logger.info(f"Scaling to {conversion_factor}")
         self._conversion_factor = conversion_factor
+        self._img = image_basics.Quick_Scale_To_im(
+            im=self._img, scale=conversion_factor
+        )
+        _logger.info("Scaled")
```
local-minimum, line 96:
```suggestion
```

local-minimum (commented):
I'm thinking that in integration tests there's a full image and perhaps it could be nice having an explicit test of this class there and the orientation marker is in `/scanomatic/images/orientation_marker.png`

scanomatic/image_analysis/first_pass_image.py
e-larsson: Hm you're right this wont work. I'll change so it crashes

e-larsson:
@local-minimum I added some integration tests to test the new functions of `FixtureImage`

PR #120, opened 2022-01-14
Add typed FixtureImage parameter

### LINKED ISSUE #118: Circular import typing `FixtureImage`
# 📋 Info

`get_grayscale` in `image_analysis.py` has a `FixtureImage` as argument but causes circular import if typed.

# 🏁 DoD

- [ ] Refactor code so it can be typed
- [ ] Consider better name for `get_grayscale` as there's another function with same name and it's a bit confusing.

e-larsson:
Not sure if I'm misunderstanding the problem but this adds the requested typehint and I don't see any circular import errors.

- Renamed `image_grayscale.py` -> `grayscale_detection.py` to make a clearer distinction from `grayscale.py`
- Renamed `get_grayscale` in `grayscale_detection` -> `detect_grayscale` to avoid confusion with `grayscale.get_grayscale`. I think this name is accurate for the function but please correct me if I'm wrong
- Rename `detect_grayscale` -> `find_grayscale_locations`. Same here as above, please correct me if I'm misunderstanding what the function does

Resolves #118

### DIFF
--- a/scanomatic/image_analysis/grayscale_detection.py
+++ b/scanomatic/image_analysis/grayscale_detection.py
@@ -11,9 +11,9 @@ from scanomatic.io.logger import get_logger
 from scanomatic.io.paths import Paths
 from scanomatic.models.fixture_models import GrayScaleAreaModel
 from scanomatic.image_analysis.first_pass_image import FixtureImage
+from scanomatic.image_analysis.grayscale import Grayscale, get_grayscale
+from scanomatic.image_analysis import signal
 
-from . import signal
-from .grayscale import Grayscale, get_grayscale
 
 ORTH_EDGE_T = 0.2
 ORTH_T1 = 0.15
--- a/scanomatic/image_analysis/grayscale_detection.py
+++ b/scanomatic/image_analysis/grayscale_detection.py
@@ -3,7 +3,6 @@ from typing import Optional
 
 import numpy as np
 from numpy.lib.stride_tricks import as_strided
-from scanomatic.models.fixture_models import GrayScaleAreaModel
 from scipy.ndimage import gaussian_filter1d  # type: ignore
 from scipy.signal import convolve2d  # type: ignore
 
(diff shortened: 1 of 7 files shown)


local-minimum:
The renaming sounds good, but could you update branch with main before review in case that triggers the circular import?

scanomatic/image_analysis/grayscale_detection.py
```diff
@@ -168,39 +169,41 @@ def get_para_trimmed_slice(
...
+        ]
+    except (IndexError, TypeError):
+        msg = "Could not extract grayscale section"
+        raise GrayscaleError(msg)
 
 
-def get_grayscale_image_analysis(
-    im: np.ndarray,
-    grayscale_name: str,
+def detect_grayscale(
```
local-minimum, line 189: Since this can now raise the calibration_api and data_api that seem to be the sources of the request should probably catch that exception and return a useful error

scanomatic/ui_server/calibration_api.py
```diff
@@ -326,9 +326,9 @@ def analyse_ccc_image_grayscale(ccc_identifier, image_identifier):
             reason='Unknown image or CCC'
         )
 
-    _, values = get_grayscale_image_analysis(
-        gs_image, gs_name, debug=False)
     grayscale_object = get_grayscale(gs_name)
+    _, values = detect_grayscale(
```
local-minimum, line 330: here's the missing try except, I think.

scanomatic/ui_server/data_api.py
```diff
@@ -292,12 +292,12 @@ def _gs_get_from_image(grayscale_name):
                 grayscale=None, reason="Area too large")
 
         try:
-            _, values = get_grayscale(
-                fixture,
+            _, values = detect_grayscale(
```
local-minimum, line 295: oh these had except already. good.

local-minimum (changes requested):

local-minimum (approved):

PR #122, opened 2022-01-17
Remove version change model

### LINKED ISSUE #92: Drop `VersionChangesModel`
# 📋 Info

It's an edge case in how data is serialized and there's probably very little need to actually support these by now very old versions of serialized data.

If we need to support that data, perhaps it is possible to migrate it by manual ways instead.

# 🏁 DoD

- [ ] Remove or restructure legacy code tied to version changes model.

local-minimum:
Move only user code to tools

Resolves #92

### DIFF
--- a/tools/image_analysis_support.py
+++ b/tools/image_analysis_support.py
@@ -1,0 +1,36 @@
+from scanomatic.generics import Model
+
+
+class VersionChangesModel(Model):
+    def __init__(self, **kwargs):
+        self.first_pass_change_1: float = 0.997
+        self.oldest_allow_fixture: float = 0.9991
+        super().__init__()
+
+
+def get_active_plates(
+    meta_data,
+    suppress_analysis,
+    graph_watch,
+    versions: VersionChangesModel,
+):
+    """Makes list of only relevant plates according to how
+    analysis was started"""
+
+    plate_position_keys = []
+
+    if meta_data['Version'] >= versions.first_pass_change_1:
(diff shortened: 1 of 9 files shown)


tools/image_analysis_support.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, **kwargs):
+        self.first_pass_change_1: float = 0.997
+        self.oldest_allow_fixture: float = 0.9991
+        super().__init__()
+
+
+def get_active_plates(
+    meta_data,
+    suppress_analysis,
+    graph_watch,
```
skymandr, line 14: I know this is dead code, but
```suggestion
    meta_data: dict,
    suppress_analysis: bool,
    graph_watch: Sequence,
```
perhaps?

skymandr (approved):
Looks good, but why are we keeping this stuff at all if we don't want to maintain it?

tools/image_analysis_support.py
```diff
@@ -0,0 +1,36 @@
+from scanomatic.generics import Model
```
local-minimum, line 1:
```suggestion
from collections.abc import Sequence

from scanomatic.generics import Model
```

local-minimum:
> Looks good, but why are we keeping this stuff at all if we don't want to maintain it?

Several functions like these may be used when working with the results in python and I'm not sure of how useful they are to current and future researchers. They code bay me unmaintained by us but it should still work.

PR #124, opened 2022-01-17
Fix queue and jobs

local-minimum:
Partial fix of queue and jobs

- Serializes down to correct form
- Can read the serialized data
- Can purge job from queue / jobs.


skymandr (approved):
lgtm

skymandr (approved):

PR #125, opened 2022-01-17
Fix some more mypy errors

e-larsson:
Part of #51

### DIFF
--- a/scanomatic/models/compile_project_model.py
+++ b/scanomatic/models/compile_project_model.py
@@ -1,5 +1,4 @@
 from enum import Enum, auto
-from typing import Optional
 from collections.abc import Sequence
 
 from scanomatic.generics.model import Model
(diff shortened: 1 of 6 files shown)


scanomatic/models/compile_project_model.py
```diff
@@ -85,11 +84,9 @@ class CompileImageAnalysisModelFields(Enum):
 class CompileImageAnalysisModel(Model):
     def __init__(
         self,
-        image: Optional[CompileImageModel] = None,
-        fixture: Optional[FixtureModel] = None,
+        image: CompileImageModel,
```
local-minimum, line 87:
```suggestion
        *,
        image: CompileImageModel,
```
Models should only accept keyword arguments. But when I re-worked this, I didn't think of that they could be keyword arguments without defaults so if this works then perhaps you could update the other models to with the asterisk.

local-minimum (changes requested):

scanomatic/models/compile_project_model.py
e-larsson: Some of the models caused errors in the factory when forcing keyword arguments, so I did this change only to the models that have falsely Optional parameters

local-minimum: and this one caused issues?

e-larsson: Oops, no this one was ok, fixed

local-minimum (approved):

PR #129, opened 2022-01-19
first upgrades

skymandr:
This is a part of #6

Here I upgrade dev-dependencies as far as I can without breaking anything. There is a lot left to do here though, as well as with the non-dev dependencies.

The most important thing in this upgrade is probably, that we can bump the node image used from 8 to 16.


package.json
```diff
@@ -2,37 +2,38 @@
...
-    "babel-plugin-istanbul": "^4.1.5",
+    "babel-plugin-istanbul": "^4.1.6",
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
-    "css-loader": "^0.28.7",
+    "css-loader": "^5.0.0",
     "enzyme": "^3.1.0",
     "enzyme-adapter-react-16": "^1.0.2",
-    "eslint": "^4.11.0",
+    "eslint": "^4.19.1",
```
skymandr, line 18: This is crucial. It could be further updated, but `5.x.x` requires an update of `eslint-config-airbnb` which in turn causes a lot of lint issues, that would make this PR very messy. I intend to try this, but would prefer to do it in a separate PR.

package.json
```diff
@@ -2,37 +2,38 @@
   "scripts": {
     "test": "BABEL_ENV=test karma start --single-run",
     "build": "webpack",
-    "lint": "eslint --ext mjs,jsx,js scanomatic/ui_server_data/js/ccc scanomatic/ui_server_data/js/specs"
+    "lint": "eslint --ext mjs,jsx,js scanomatic/ui_server_data/js/ccc scanomatic/ui_server_data/js/tests"
```
skymandr, line 5: Oops!

local-minimum (commented):
Not much to say but I guessed you've brought up SoM and tested interacting with the various views to see that everything looks 🆗

skymandr:
> Not much to say but I guessed you've brought up SoM and tested interacting with the various views to see that everything looks ok

Yes, and as far as I can tell it looks ok. It will be more annoying to check when we get to the actual deps...

local-minimum (approved):

PR #131, opened 2022-01-19
Fix bugs encountered when running SoM

local-minimum:
- Handle error when requesting status
- Handle not having any images in analysis (Happens when attempting
to run old project)
- Fix json abort that was missing status code


skymandr (approved):
lgtm

PR #135, opened 2022-01-20
Create codeql-analysis.yml

skymandr:
This is a test to see if Code QL can give us anything.


local-minimum (approved):
Looks terrifying but great tool

Issue #47, opened 2021-10-29
Validate can do QC


local-minimum:
# 📋 Info
**Partially blocked** by #46 

Do various QC tasks on project

# 🏁 DoD

- [x] New project from #46
- [x] Old project from python 2.7 times

skymandr:
Found when loading a complete converted project in #99:
```console
ERROR	**Scan-o-Matic UI** Exception on /api/results/phenotype_names/PM2/PayamMartin19Oct16Scanner2/analysis_original [GET]

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1518, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.9/site-packages/flask_restful/__init__.py", line 271, in error_router
    return original_handler(e)
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/usr/local/lib/python3.9/site-packages/scanomatic/ui_server/qc_api.py", line 705, in get_phenotype_names
    state, _ = _get_state_update_response(path, response, success=True)
  File "/usr/local/lib/python3.9/site-packages/scanomatic/ui_server/qc_api.py", line 69, in _get_state_update_response
    state = phenotyper.Phenotyper.LoadFromState(path)
  File "/usr/local/lib/python3.9/site-packages/scanomatic/data_processing/phenotyper.py", line 354, in LoadFromState
    PhenotypeDataType[inclusion_name],
  File "/usr/local/lib/python3.9/enum.py", line 432, in __getitem__
    return cls._member_map_[name]
KeyError: b'Trusted'
```

local-minimum:
The issue there seems to be that it is loaded as bytes while since it should load as an enum value must be string format.

PR #136, opened 2022-01-21
Issue 47 validate qc part 1

skymandr:
First of probably many parts of #47. Also brings number of mypy errors below 500! 🥳 

What works:
- can load a converted project (using https://github.com/Scan-o-Matic/project-converter/pull/4)
- can interact with plates, see different phenotypes, and mark colonies
- can inspect grid
- can run normalize

What doesn't
- [ ] there are tonnes of console errors from mouse events when interacting with the plate (something with `d3.event`)
- [x] ~the `scan.instructions` can't be loaded, which causes some bugs~ (fixed by https://github.com/Scan-o-Matic/project-converter/pull/5)
- [x] ~jsonizer not fully integrated~ (fixed
- [ ] though normalization can be run, I'm not sure it does anything reasonable, and interacting with the normalized phenotypes causes a hang)

In addition the following warnings (and probably others) end up in the docker-compose log:
```console
WARNING	**Phenotyper** There are gaps in the time series at times: 42.738247944444446 - 43.406780194444444, 69.11481372222222 - 69.78131019444444
WARNING	**Phenotyper** Outdated phenotypes format, guessing which were extracted and converting to current format.To faster load in the future and get rid of this message, save current state.
WARNING	**Phenotyper** The plates don't agree on what phenotypes were extracted, you should tread carefully or better yet: rerun feature extraction.
WARNING	**Phenotyper State** Filter doesn't match number of plates. Rewriting...
```

### DIFF
--- a/scanomatic/data_processing/pheno/save.py
+++ b/scanomatic/data_processing/pheno/save.py
@@ -1,3 +1,4 @@
+import json
 import os
 import pickle
 import zipfile
@@ -7,6 +8,7 @@ from typing import Optional
 import numpy as np
 
 import scanomatic.io.paths as paths
+import scanomatic.io.jsonizer as jsonizer
 from scanomatic.data_processing.pheno.state import (
     PhenotyperSettings,
     PhenotyperState
@@ -127,7 +129,10 @@ def save_state(
         or not os.path.isfile(p)
         or _do_ask_overwrite(p)
     ):
-        np.save(p, settings.serialized())
+        with open(p, 'w') as fp:
+            json.dump(json.loads(
+                jsonizer.dumps(settings.asdict()),
+            ), fp)
 
     _logger.info("State saved to '{0}'".format(dir_path))
 
--- a/scanomatic/data_processing/pheno/save.py
+++ b/scanomatic/data_processing/pheno/save.py
@@ -140,12 +140,11 @@ def save_state_to_zip(
     state: PhenotyperState,
     target: Optional[str] = None,
 ) -> Optional[BytesIO]:
-    JsonizerStateWriter = Callable[[BufferedWriter, Any], int]
-    PickleStateWriter = Callable[[BufferedWriter, Any], None]
-    save_jsonizer: JsonizerStateWriter = (
+    StateWriter = Callable[[BufferedWriter, Any], Any]
+    save_jsonizer: StateWriter = (
         lambda fh, obj: fh.write(jsonizer.dumps(obj).encode())
     )
-    save_pickle: PickleStateWriter = lambda fh, obj: pickle.dump(obj, fh)
+    save_pickle: StateWriter = lambda fh, obj: pickle.dump(obj, fh)
 
     def zipit(save_functions, data, zip_paths):
         zip_buffer = BytesIO()
@@ -181,9 +180,9 @@ def save_state_to_zip(
     if not dir_path or not dir_path.strip() or dir_path == ".":
         dir_path = "analysis"
 
-    save_functions: list[Callable] = []
+    save_functions: list[StateWriter] = []
     data: list[Any] = []
-    zip_paths = []
+    zip_paths: list[str] = []
 
     # Phenotypes
     zip_paths.append(
--- a/scanomatic/data_processing/pheno/save.py
+++ b/scanomatic/data_processing/pheno/save.py
@@ -1,12 +1,11 @@
-import json
 import os
 import pickle
 import zipfile
-from io import BytesIO
-from typing import Optional
+from collections.abc import Callable
+from io import BufferedWriter, BytesIO
+from typing import Any, Optional
 
 import numpy as np
-from dataclasses import asdict
 
 import scanomatic.io.paths as paths
 import scanomatic.io.jsonizer as jsonizer
@@ -130,10 +129,7 @@ def save_state(
         or not os.path.isfile(p)
         or _do_ask_overwrite(p)
     ):
-        with open(p, 'w') as fp:
-            json.dump(json.loads(
-                jsonizer.dumps(asdict(settings)),
-            ), fp)
+        jsonizer.dump(settings, p)
 
     _logger.info("State saved to '{0}'".format(dir_path))
 
@@ -144,6 +140,13 @@ def save_state_to_zip(
     state: PhenotyperState,
     target: Optional[str] = None,
 ) -> Optional[BytesIO]:
+    JsonizerStateWriter = Callable[[BufferedWriter, Any], int]
+    PickleStateWriter = Callable[[BufferedWriter, Any], None]
+    save_jsonizer: JsonizerStateWriter = (
+        lambda fh, obj: fh.write(jsonizer.dumps(obj).encode())
+    )
+    save_pickle: PickleStateWriter = lambda fh, obj: pickle.dump(obj, fh)
+
     def zipit(save_functions, data, zip_paths):
         zip_buffer = BytesIO()
         zf = zipfile.ZipFile(zip_buffer, 'a', zipfile.ZIP_DEFLATED, False)
@@ -178,8 +181,8 @@ def save_state_to_zip(
     if not dir_path or not dir_path.strip() or dir_path == ".":
         dir_path = "analysis"
 
-    save_functions = []
-    data = []
+    save_functions: list[Callable] = []
+    data: list[Any] = []
     zip_paths = []
 
     # Phenotypes
@@ -242,7 +245,7 @@ def save_state_to_zip(
     zip_paths.append(
         os.path.join(dir_path, _paths.phenotypes_filter_undo),
     )
-    save_functions.append(lambda fh, obj: pickle.dump(obj, fh))
+    save_functions.append(save_pickle)
     data.append(state.phenotype_filter_undo)
 
     # Time stamps
@@ -256,15 +259,15 @@ def save_state_to_zip(
     zip_paths.append(
         os.path.join(dir_path, _paths.phenotypes_meta_data),
     )
-    save_functions.append(lambda fh, obj: pickle.dump(obj, fh))
+    save_functions.append(save_pickle)
     data.append(state.meta_data)
 
     # Internal settings
     zip_paths.append(
         os.path.join(dir_path, _paths.phenotypes_extraction_params),
     )
-    save_functions.append(lambda fh, obj: json.dump(obj, fh))
-    data.append(json.loads(jsonizer.dumps(settings.asdict())))
+    save_functions.append(save_jsonizer)
+    data.append(settings)
 
     zip_stream = zipit(save_functions, data, zip_paths)
     if target:
@@ -272,5 +275,6 @@ def save_state_to_zip(
             fh.write(zip_stream.read())
         zip_stream.close()
         _logger.info("Zip file saved to {0}".format(target))
+        return None
     else:
         return zip_stream
--- a/scanomatic/data_processing/pheno/save.py
+++ b/scanomatic/data_processing/pheno/save.py
@@ -6,6 +6,7 @@ from io import BytesIO
 from typing import Optional
 
 import numpy as np
+from dataclasses import asdict
 
 import scanomatic.io.paths as paths
 import scanomatic.io.jsonizer as jsonizer
@@ -131,7 +132,7 @@ def save_state(
     ):
         with open(p, 'w') as fp:
             json.dump(json.loads(
-                jsonizer.dumps(settings.asdict()),
+                jsonizer.dumps(asdict(settings)),
             ), fp)
 
     _logger.info("State saved to '{0}'".format(dir_path))
--- a/scanomatic/data_processing/pheno/save.py
+++ b/scanomatic/data_processing/pheno/save.py
@@ -262,8 +262,8 @@ def save_state_to_zip(
     zip_paths.append(
         os.path.join(dir_path, _paths.phenotypes_extraction_params),
     )
-    save_functions.append(np.save)
-    data.append(settings.serialized())
+    save_functions.append(lambda fh, obj: json.dump(obj, fh))
+    data.append(json.loads(jsonizer.dumps(settings.asdict())))
 
     zip_stream = zipit(save_functions, data, zip_paths)
     if target:
--- a/scanomatic/data_processing/pheno/state.py
+++ b/scanomatic/data_processing/pheno/state.py
@@ -1,5 +1,5 @@
 from collections import deque
-from dataclasses import dataclass, field
+from dataclasses import asdict, dataclass, field
 from typing import Any, Optional, Union
 
 import numpy as np
@@ -35,8 +35,8 @@ class PhenotyperSettings:
     gaussian_filter_sigma: float
     linear_regression_size: int
     phenotypes_inclusion: Optional[PhenotypeDataType]
-    no_growth_monotonicity_threshold: float
-    no_growth_pop_doublings_threshold: float
+    no_growth_monotonicity_threshold: Optional[float]
+    no_growth_pop_doublings_threshold: Optional[float]
 
     def __post_init__(self):
         assert (
@@ -50,9 +50,14 @@ class PhenotyperSettings:
             self.linear_regression_size,
             None if self.phenotypes_inclusion is None
             else self.phenotypes_inclusion.name,
-            self.no_growth_monotonicity_threshold,
-            self.no_growth_pop_doublings_threshold
+            None if self.no_growth_monotonicity_threshold is None
+            else self.no_growth_monotonicity_threshold,
+            None if self.no_growth_pop_doublings_threshold is None
+            else self.no_growth_pop_doublings_threshold,
         ]
+
+    def asdict(self) -> dict[str, Any]:
+        return asdict(self)
 
 
 @dataclass
--- a/scanomatic/data_processing/pheno/state.py
+++ b/scanomatic/data_processing/pheno/state.py
@@ -1,5 +1,5 @@
 from collections import deque
-from dataclasses import asdict, dataclass, field
+from dataclasses import dataclass, field
 from typing import Any, Optional, Union
 
 import numpy as np
@@ -42,22 +42,6 @@ class PhenotyperSettings:
         assert (
             self.median_kernel_size % 2 == 1
         ), "Median kernel size must be odd"
-
-    def serialized(self) -> list[Union[str, int, float]]:
-        return [
-            self.median_kernel_size,
-            self.gaussian_filter_sigma,
-            self.linear_regression_size,
-            None if self.phenotypes_inclusion is None
-            else self.phenotypes_inclusion.name,
-            None if self.no_growth_monotonicity_threshold is None
-            else self.no_growth_monotonicity_threshold,
-            None if self.no_growth_pop_doublings_threshold is None
-            else self.no_growth_pop_doublings_threshold,
-        ]
-
-    def asdict(self) -> dict[str, Any]:
-        return asdict(self)
 
 
 @dataclass
--- a/scanomatic/data_processing/pheno/state.py
+++ b/scanomatic/data_processing/pheno/state.py
@@ -34,9 +34,11 @@ class PhenotyperSettings:
     median_kernel_size: int
     gaussian_filter_sigma: float
     linear_regression_size: int
-    phenotypes_inclusion: Optional[PhenotypeDataType]
-    no_growth_monotonicity_threshold: Optional[float]
-    no_growth_pop_doublings_threshold: Optional[float]
+    phenotypes_inclusion: Optional[PhenotypeDataType] = (
+        PhenotypeDataType.Trusted
+    )
+    no_growth_monotonicity_threshold: Optional[float] = None
+    no_growth_pop_doublings_threshold: Optional[float] = None
 
     def __post_init__(self):
         assert (
--- a/scanomatic/data_processing/phenotyper.py
+++ b/scanomatic/data_processing/phenotyper.py
@@ -332,7 +332,6 @@ class Phenotyper(mock_numpy_interface.NumpyArrayInterface):
         )
 
         try:
-            
             extraction_params = jsonizer.load(os.path.join(
                 directory_path,
                 _p.phenotypes_extraction_params,
--- a/scanomatic/data_processing/phenotyper.py
+++ b/scanomatic/data_processing/phenotyper.py
@@ -353,10 +353,9 @@ class Phenotyper(mock_numpy_interface.NumpyArrayInterface):
             )
         else:
             try:
-                phenotype_inclusion = extraction_params.get(
+                if (phenotype_inclusion := extraction_params.get(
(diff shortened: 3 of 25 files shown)


scanomatic/data_processing/pheno/save.py
```diff
@@ -127,7 +129,10 @@ def save_state(
         or not os.path.isfile(p)
         or _do_ask_overwrite(p)
     ):
-        np.save(p, settings.serialized())
+        with open(p, 'w') as fp:
+            json.dump(json.loads(
+                jsonizer.dumps(settings.asdict()),
+            ), fp)
```
skymandr, line 135: I'm not proud of this, but not sure in what way it would be reasonable to generalize the jsonizer... Suggestions welcome!

local-minimum:
>     * [ ]  the `scan.instructions` can't be loaded, which causes some bugs

Why not just adding it in the converter PR, it should be much easier than what you did with for the other serialized state

scanomatic/data_processing/pheno/save.py
local-minimum: `jsonizer` has a `dump` too. also I don't know if we need an extra asdict on settings, cant we use asdict imported from dataclasses directly? Finally I think it would be more in line with the intent of the jsonizer if convertion from settings to dict happened inside jsonizer.dumps just like the enums and models are treated.

scanomatic/data_processing/pheno/state.py
```diff
@@ -50,10 +50,15 @@ def serialized(self) -> list[Union[str, int, float]]:
...
-            self.no_growth_monotonicity_threshold,
-            self.no_growth_pop_doublings_threshold
+            None if self.no_growth_monotonicity_threshold is None
+            else self.no_growth_monotonicity_threshold,
+            None if self.no_growth_pop_doublings_threshold is None
+            else self.no_growth_pop_doublings_threshold,
         ]
 
+    def asdict(self) -> dict[str, Any]:
+        return asdict(self)
```
local-minimum, line 60: this is useless, remove

scanomatic/data_processing/pheno/state.py
```diff
@@ -50,10 +50,15 @@ def serialized(self) -> list[Union[str, int, float]]:
             self.linear_regression_size,
             None if self.phenotypes_inclusion is None
             else self.phenotypes_inclusion.name,
-            self.no_growth_monotonicity_threshold,
-            self.no_growth_pop_doublings_threshold
+            None if self.no_growth_monotonicity_threshold is None
+            else self.no_growth_monotonicity_threshold,
+            None if self.no_growth_pop_doublings_threshold is None
+            else self.no_growth_pop_doublings_threshold,
```
local-minimum, line 56: what does this do? Also is serialized really used now, can't it be removed.

scanomatic/data_processing/phenotyper.py
```diff
@@ -322,73 +328,62 @@ def LoadFromState(cls, directory_path: str) -> "Phenotyper":
...
-                        linear_reg_size,
-                        inclusion_name,
-                    ) = extraction_params
-                    if inclusion_name is None:
-                        inclusion_name = 'Trusted'
+            try:
+                phenotype_inclusion = extraction_params.get(
+                    "phenotype_inclusion",
+                )
+                if phenotype_inclusion is not None:
```
local-minimum, line 348: perhaps rewrite with `:=`

scanomatic/data_processing/phenotyper.py
```diff
@@ -322,73 +328,62 @@ def LoadFromState(cls, directory_path: str) -> "Phenotyper":
...
-                        inclusion_name = 'Trusted'
-
+                else:
+                    phenotyper.set_phenotype_inclusion_level(
+                        PhenotypeDataType.Trusted,
+                    )
+                no_growth_monotonicity_threshold = extraction_params.get(
+                    "no_growth_monotonicity_threshold",
+                )
+                if no_growth_monotonicity_threshold is not None:
```
local-minimum, line 359: same

scanomatic/data_processing/phenotyper.py
```diff
@@ -322,73 +328,62 @@ def LoadFromState(cls, directory_path: str) -> "Phenotyper":
...
                     phenotyper._settings.no_growth_monotonicity_threshold = (
-                        float(no_growth_monotonicity_threshold)
+                        no_growth_monotonicity_threshold
                     )
+                no_growth_pop_doublings_threshold = extraction_params.get(
+                    "no_growth_pop_doublings_threshold",
+                )
+                if no_growth_pop_doublings_threshold is not None:
```
local-minimum, line 366: same

scanomatic/data_processing/phenotyper.py
```diff
@@ -322,73 +328,62 @@ def LoadFromState(cls, directory_path: str) -> "Phenotyper":
...
             os.path.join(directory_path, _p.phenotypes_input_smooth),
+            allow_pickle=True,
         )
 
         try:
-            extraction_params = unpickle_with_unpickler(
-                np.load,
-                os.path.join(directory_path, _p.phenotypes_extraction_params),
-            )
+            extraction_params = jsonizer.load(os.path.join(
```
local-minimum, line 335: What type is this? It should be the dataclass it used to be, but it looks like it is now a dict and if it is, I think it will be quite problematic. If it is a dict I would argue there should be a new jsonizer serializer/deserializer for dataclasses just like enums and Model:s

scanomatic/data_processing/phenotyper.py
```diff
@@ -537,8 +544,16 @@ def LoadFromNumPy(
...
-            unpickle_with_unpickler(np.load, times_data_path),
+            unpickle_with_unpickler(
+                np.load,
+                data_directory,
+                allow_pickle=True,
+            ),
+            unpickle_with_unpickler(
+                np.load,
+                times_data_path,
+                allow_pickle=True,
```
local-minimum, line 555: Are all uses of this function now calling it with allow_pickle=True? If so consider removing the argument and simplifying that function.

scanomatic/data_processing/project.py
```diff
@@ -32,19 +34,23 @@ def path_has_saved_project_state(
...
+            allow_pickle=True,
         )
-        unpickle_with_unpickler(
-            np.load,
+        with open(
             os.path.join(directory_path, _p.phenotypes_extraction_params),
-        )
+            'r',
+        ) as fp:
+            json.load(fp)
```
local-minimum, line 53: this ought to be jsonizer and not just json

local-minimum (changes requested):

scanomatic/data_processing/pheno/state.py
skymandr: Yes, it can safely be removed.

skymandr:
> > ```
> > * [ ]  the `scan.instructions` can't be loaded, which causes some bugs
> > ```
> 
>Why not just adding it in the converter PR, it should be much easier than what you did with for the other serialized state

I have started on a PR for that, so far without success. I think this PR is messy enough without it.

scanomatic/data_processing/pheno/state.py
skymandr: I tried using the `jsonizerd.dump`, but how the jsonizer needs to be extended/generalized to accommodate this type of object seemed to be non-trivial. I agree it should be done, but I would like to discuss how, before I set off in the wrong direction.

scanomatic/data_processing/phenotyper.py
skymandr: The thought struck me, but everytime I've tried being clever with the `:=` and `is None` it has backfired on some technicality. I can have another look, but don't hold your breath.

scanomatic/data_processing/phenotyper.py
skymandr: No, previously it was just an array, not a dataclass.

scanomatic/data_processing/phenotyper.py
skymandr: `unpickle_with_unpickler` is not, or is atleast not meant to, be called with `np.load`, to which `allow_pickle` is an argument. I'll have a look. If it is only used with `np.load` I guess I can remove that argument too.

scanomatic/data_processing/project.py
skymandr: Agreed, but see above.

scanomatic/data_processing/phenotyper.py
local-minimum: maybe missing parenthesis

scanomatic/data_processing/phenotyper.py
local-minimum: When it is created in the app before serialization it is a dataclass so it should be that when loaded

scanomatic/data_processing/pheno/save.py
skymandr: I tried using the jsonizerd.dump, but how the jsonizer needs to be extended/generalized to accommodate this type of object seemed to be non-trivial. I agree it should be done, but I would like to discuss how, before I set off in the wrong direction. It only expects `Models` at the moment.

scanomatic/data_processing/phenotyper.py
skymandr: Yes, this whole method is about creating that dataclass from the json.

scanomatic/data_processing/phenotyper.py
skymandr: As far as I can tell, it is used exclusively with `np.load`. One way forward is to replace each call to `unpickle_with_unpickler` with 
```python
np.load(
    safe_load(p),
    allow_pickle=True,
)
```
and remove `unpickle_with_unpickler` entirely. Either that, or renaming it `safe_load_npy` and doing precisely that, but save on arguments in the bulk of the code. What do you think?

local-minimum: if that  is all that it does now, then I'm all for removing the unpickle method and just use np.load in code

scanomatic/data_processing/phenotyper.py
skymandr: fixed

scanomatic/data_processing/phenotyper.py
skymandr: fixed

scanomatic/data_processing/phenotyper.py
skymandr: fixed

scanomatic/data_processing/pheno/state.py
skymandr: fixed

scanomatic/data_processing/pheno/state.py
skymandr: fixed

scanomatic/data_processing/phenotyper.py
skymandr: fixed

scanomatic/data_processing/project.py
skymandr: fixed

scanomatic/data_processing/pheno/save.py
local-minimum: dump(s) should perhaps rightly accept `Any`

scanomatic/data_processing/pheno/save.py
```diff
@@ -127,7 +130,10 @@ def save_state(
         or not os.path.isfile(p)
         or _do_ask_overwrite(p)
     ):
-        np.save(p, settings.serialized())
+        with open(p, 'w') as fp:
+            json.dump(json.loads(
+                jsonizer.dumps(asdict(settings)),
+            ), fp)
```
local-minimum, line 136: Should only be this
```suggestion
        jsonizer.dumps(settings, p)
```

scanomatic/data_processing/pheno/save.py
```diff
@@ -257,8 +263,8 @@ def zipit(save_functions, data, zip_paths):
     zip_paths.append(
         os.path.join(dir_path, _paths.phenotypes_extraction_params),
     )
-    save_functions.append(np.save)
-    data.append(settings.serialized())
+    save_functions.append(lambda fh, obj: json.dump(obj, fh))
+    data.append(json.loads(jsonizer.dumps(settings.asdict())))
```
local-minimum, line 267: Should be something like this
```suggestion
    save_functions.append(lambda fh, obj: fh.write(jsonizer.dumps(obj)))
    data.append(settings)
```

scanomatic/data_processing/phenotyper.py
```diff
@@ -269,14 +269,20 @@ def LoadFromState(cls, directory_path: str) -> "Phenotyper":
         """
         _p = paths.Paths()
 
-        raw_growth_data = unpickle_with_unpickler(np.load, os.path.join(
-            directory_path,
-            _p.phenotypes_input_data
-        ))
+        raw_growth_data = np.load(
+            safe_load(os.path.join(
```
local-minimum, line 273: what does safe_load to / is it needed?

scanomatic/data_processing/phenotyper.py
local-minimum: no it's about loading a phenotyper state, but the settings should just be `settings: PhenotyperSettings = jsonizer.load(the_path)`

local-minimum: most, if not all, from this line to line 395 should be removed (and should not be needed in the deserializer in jsonizer

scanomatic/io/pickler.py
```diff
@@ -20,25 +20,6 @@ def safe_load(path, return_string=False):
...
-
-    Args:
-        unpickler (func): A function that unpickles a string of data
-        path (str): Path to the data to be unpickled
-        *args: Optional extra arguments for the unpickler
-        **kwargs: Opional extra keyword arguments for the unpickler
-
-    Returns: Unpickled object
-    """
-    return unpickler(safe_load(path), *args, **kwargs)
```
local-minimum, line 39: is unpickler used now?

local-minimum (changes requested):

scanomatic/data_processing/phenotyper.py
skymandr: It that would work, that would be lovely. Bearing that in mind though, you did tell me to take short cuts, and I think this, besides the fact that it works, is already an improvement on how it used to work. Please make an executive call on what you think is appropriate.

scanomatic/data_processing/phenotyper.py
skymandr: It does:
```python
def safe_load(path, return_string=False):

    version_compatibility = _RefactoringPhases()
    fh = SafeProxyFileObject(path, version_compatibility)
    version_compatibility.io = fh
    if return_string:
        return os.linesep.join(fh.readlines())
    else:
        return fh
 ```
 Whether it is needed or not I can scarcely guess at, but it seems to be doing something for compatibility..?

File: scanomatic/data_processing/pheno/save.py (hunk 10 above)
Comment (skymandr) [reply scanomatic/data_processing/pheno/save.py:136]: I assume you mean `jsonizer.dump`. I'll see if I can get that to work, it would be much neater. Should it have the default value `False` on `overwrite` though?

File: scanomatic/data_processing/pheno/save.py (hunk 11 above)
Comment (skymandr) [reply scanomatic/data_processing/pheno/save.py:267]: I guess the `SOMSerializers` and `SOMEncoder` needs to be expanded for that to work..?

File: scanomatic/io/pickler.py (hunk 13 above)
Comment (skymandr) [reply scanomatic/io/pickler.py:39]: No.

Comment (skymandr) [reply scanomatic/io/pickler.py:39]: Or well, `unpickler` was an argument to `unpickle_with_unpickler`, but I assume you mean `unpickle`, which is still used, though only in two places.

Comment (skymandr) [reply scanomatic/io/pickler.py:39]: `unpickle` only does this:
```def unpickle(path: str):
    """Unpickles data safely
    Returns: Unpickled object
    """
    return pickle.load(safe_load(path))
```

scanomatic/data_processing/pheno/save.py
local-minimum: overwrite is an old name and a bit confusing. it really means merge into

scanomatic/data_processing/pheno/save.py
local-minimum: yes SOMEncoder only needs an `isinstance(o, dataclass)` and then `asdict(o)` in it

local-minimum: and the `decode_dataclass` should be fairly straight forward. Just a `lookup[name](**obj[CONTENT])`

scanomatic/data_processing/pheno/save.py
skymandr: and it should be `False`?

scanomatic/data_processing/pheno/state.py
```diff
@@ -35,25 +35,14 @@ class PhenotyperSettings:
     gaussian_filter_sigma: float
     linear_regression_size: int
     phenotypes_inclusion: Optional[PhenotypeDataType]
-    no_growth_monotonicity_threshold: float
-    no_growth_pop_doublings_threshold: float
+    no_growth_monotonicity_threshold: Optional[float]
+    no_growth_pop_doublings_threshold: Optional[float]
```
local-minimum, line 39: these should have default `None` right? and `phenotypes_inclusion` shouldn't really be optional but have a default value too right?

scanomatic/data_processing/phenotyper.py
local-minimum: might might be needed, so lets leave it for now though

scanomatic/data_processing/phenotyper.py
local-minimum: I think jsonizer should load it

scanomatic/data_processing/pheno/state.py
skymandr: `phenotypes_inclusion` was strictly optional in the saved settings, but it should default to `PhenotypeDataType.Trusted`.

scanomatic/io/pickler.py
local-minimum: yea probably nicer to inline

scanomatic/data_processing/pheno/save.py
local-minimum: yes I belive so, False means it just dumps the content to the target location independent of what is there, 'overwrite' means it tries to merge new info with existing info...

skymandr: I'll rename it as well then.

scanomatic/data_processing/pheno/state.py
local-minimum: so better `no_growth_pop_doublings_threshold: Optional[float] = None` etc

scanomatic/io/pickler.py
skymandr: fixed (next push)

scanomatic/data_processing/phenotyper.py
skymandr: fixed

scanomatic/data_processing/pheno/save.py
skymandr: fixed

scanomatic/data_processing/pheno/save.py
skymandr: fixed

skymandr:
Please see updated PR description for notes!

scanomatic/data_processing/pheno/save.py
```diff
@@ -138,6 +140,13 @@ def save_state_to_zip(
     state: PhenotyperState,
     target: Optional[str] = None,
 ) -> Optional[BytesIO]:
+    JsonizerStateWriter = Callable[[BufferedWriter, Any], int]
+    PickleStateWriter = Callable[[BufferedWriter, Any], None]
```
local-minimum, line 144: Since i guess we don't really care about return values, they just happen to be something, then maybe this could be a single `Callable[[BufferedWriter, Any], Any]`? Then it could be used for the `save_functions` variable below too.

scanomatic/io/jsonizer.py
```diff
@@ -185,6 +188,33 @@ def decode_array(obj: dict) -> np.ndarray:
         raise JSONDecodingError(msg)
 
 
+DATA_CLASSES: dict[str, Any] = {
+    "PhenotyperSettings": PhenotyperSettings,
+}
+
+
+def decode_dataclass(obj: dict) -> Any:
```
local-minimum, line 196: nice, exactly like I had in mind

scanomatic/io/jsonizer.py
```diff
@@ -229,6 +260,15 @@ def default(self, o: Any) -> Any:
                 SOMSerializers.ARRAY.encoding: o.dtype.name,
                 CONTENT: o.tolist()
             }
+        elif is_dataclass(o):
```
local-minimum, line 263: missing test cases for this new feature

scanomatic/data_processing/pheno/save.py
skymandr: fixed

scanomatic/io/jsonizer.py
```diff
@@ -328,11 +368,11 @@ def merge_into(
 
 
 def dump(
-    model: Union[Model, Sequence[Model]],
+    model: Any,
```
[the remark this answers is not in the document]
skymandr: Mypy doesn't actually complain about this, but it makes sense to update.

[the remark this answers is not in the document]
skymandr: fixed

scanomatic/io/jsonizer.py
skymandr: fixed, I think

scanomatic/io/jsonizer.py
skymandr: fixed, I think

tests/unit/io/fixtures/phenotype_params.json
```diff
@@ -0,0 +1,14 @@
...
+  "__CONTENT__": {
+    "no_growth_pop_doublings_threshold": 1.0,
+    "phenotypes_inclusion": {
+      "__CONTENT__": "Trusted",
+      "__ENUM__": "PhenotypeDataType"
+    },
+    "median_kernel_size": 5,
+    "linear_regression_size": 5,
+    "no_growth_monotonicity_threshold": 0.6,
+    "gaussian_filter_sigma": 1.5
```
local-minimum, line 12: Perhaps it would be nicer if there was a version with only mandatory fields too.

local-minimum (commented):
You should add a test-case to `def test_preserves_model(model: Model):` too so that the dumping is properly tested too

tests/unit/io/fixtures/phenotype_params.json
skymandr: fixed (this and main comment)

local-minimum (approved):

Issue #132, opened 2022-01-19
Validate can run analysis from compiled project


local-minimum:
# 📋 Info

Analysis should be able to run from compiled projects

# 🏁 DoD

- [ ] Outputs a project that can be used to extract features

Issue #33, opened 2021-10-27
New Grayscale detection algorithm


local-minimum:
# 📋 Info
General problems with identifying grayscale, selecting grayscale when identifying a fixture

We need to find a more durable option for gray scale. The current method allows it to accumulate dust and degrade rather quickly. Moreover, if somehow the grayscale moved / tilted slightly in its place, it becomes impossible to analyse that project data without gray scale hacking. 

# Requirements
- As many as images possible with troublesome GS detection from as many different projects and fixtures as posible
- As many as images possible with deteriorated GS from as many different projects and fixtures as posible

# 🏁 DoD

- [ ] Implement more robust GS detection

dodslaser:
Grayscale detection has been a major pain-point for us at GU, to the point where people are habitually hacking old grayscale values into their scans.

I know this has been brought up in the past, but we would like to have to bypass this somehow. One option would be to run calibration routinely (e.g. on a schedule, or before each project) and then use those values, as this is essentially what we have been doing anyway. As far as we can tell, the effect is minimal. Preferably, we would like to run calibration on an unmodified IT8 target of some format.

local-minimum:
I agree, the main need for this calibration was when the scanners didn't use LED lamps. The LED light source seems much more stable.

It could be possible to make a "no-grascale" grayscale detection that just outputs the same values no matter the image and then there's no need to touch the other parts of the project since they will believe that they detected a grayscale like they've always done.

PR #140, opened 2022-01-26
Validate compile project

### LINKED ISSUE #46: Validate can compile project from images
# 📋 Info

Use real images of a project but no prior analysis

# 🏁 DoD

- [x] Can make project / first pass analysis
- [x] Can run analysis on resulting compilation

e-larsson:
Resolves #46 and #132 (I think?)

- [x] Compile project works using both local and global fixture configs in json format 
The global fixture config file **must** be named "almost" like its `name` field, e.g if the name field states `Delta 1`, the file must be named `delta_1.config`
- [x] Analysis works with default settings
I'm not able to get the "Manual regridding of previously analysed project" to be activated, but I don't see any errors, so not sure if I'm missing something here

### DIFF
--- a/scanomatic/image_analysis/first_pass_image.py
+++ b/scanomatic/image_analysis/first_pass_image.py
@@ -433,7 +433,7 @@ class FixtureImage:
             current_model.grayscale.section_values = detect_grayscale(
                 grayscale_im,
                 grayscale_config,
-            )[1]
+            )
 
         except GrayscaleError:
             self._logger.warning("Failed analysing grayscale")
--- a/scanomatic/image_analysis/first_pass_image.py
+++ b/scanomatic/image_analysis/first_pass_image.py
@@ -425,7 +425,10 @@ class FixtureImage:
     def analyse_grayscale(self) -> None:
         current_model = self["current"].model
         try:
-            grayscale_im = get_grayscale_im_section(self.im, current_model.grayscale)
+            grayscale_im = get_grayscale_im_section(
+                self.im,
+                current_model.grayscale,
+            )
             grayscale_config = get_grayscale(current_model.grayscale.name)
             current_model.grayscale.section_values = detect_grayscale(
                 grayscale_im,
--- a/scanomatic/image_analysis/first_pass_image.py
+++ b/scanomatic/image_analysis/first_pass_image.py
@@ -5,6 +5,7 @@ from typing import Any, Union
 import numpy as np
 
 from scanomatic.data_processing.calibration import get_image_json_from_ccc
+from scanomatic.image_analysis.grayscale import get_grayscale
 from scanomatic.io.ccc_data import CCCImage
 from scanomatic.io.fixtures import Fixtures, FixtureSettings
 from scanomatic.io.logger import get_logger
@@ -15,6 +16,7 @@ from scanomatic.models.fixture_models import (
 )
 from scanomatic.image_analysis.grayscale_detection import (
     detect_grayscale,
+    get_grayscale_im_section,
 )
 from scanomatic.image_analysis.exceptions import GrayscaleError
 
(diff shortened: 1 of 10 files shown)


scanomatic/image_analysis/first_pass_image.py
```diff
@@ -423,9 +425,14 @@ def get_plate_im_section(
     def analyse_grayscale(self) -> None:
         current_model = self["current"].model
         try:
-            current_model.grayscale.section_values = detect_grayscale(
+            grayscale_im = get_grayscale_im_section(
                 self.im,
                 current_model.grayscale,
+            )
+            grayscale_config = get_grayscale(current_model.grayscale.name)
+            current_model.grayscale.section_values = detect_grayscale(
```
local-minimum, line 433: just a curious question, we're only saving the `1` index, is this always the case from detect_grayscale? In that case perhaps nice to simplify the return type.

scanomatic/image_analysis/grid_cell_extra.py
```diff
@@ -602,7 +602,7 @@ def detect(
             if self.filter_array.sum() == 0:
                 self.filter_array = self.old_filter.copy()
 
-            blob_diff = (np.abs(self.old_filter - self.filter_array)).sum()
+            blob_diff = (np.abs(self.old_filter ^ self.filter_array)).sum()
```
local-minimum, line 605: can you explain this and the following changes, why they are valid.

scanomatic/ui_server/data_api.py
```diff
@@ -362,22 +364,25 @@ def _gs_get_from_fixture(fixture_name):
...
-
+        grayscale_config = get_grayscale(grayscale_area_model.name)
         try:
+            grayscale_im = get_grayscale_im_section(
+                fixture.im,
+                grayscale_config,
+            )
             _, values = detect_grayscale(
-                fixture.im, grayscale_area_model, debug=is_debug_mode)
+                grayscale_im, grayscale_config, debug=is_debug_mode)
```
local-minimum, line 374:
```suggestion
                grayscale_im, grayscale_config, debug=is_debug_mode,
           )
```

local-minimum (commented):
In general looking good, just some questions

scanomatic/image_analysis/first_pass_image.py
e-larsson: Yes! It is always the case. But it is however a bit strange, it means that we are not using the `grayscale_segment_centers` from `grayscale_detection.find_grayscale_locations` anywhere in the code. That seems odd?

scanomatic/image_analysis/grid_cell_extra.py
e-larsson: `self.old_filter` and `self.filter_array` are numpy boolean arrays. Doing the subtraction results in the following error
```
TypeError: numpy boolean subtract, the `-` operator, is not supported, use the bitwise_xor, the `^` operator, or the logical_xor function instead.
```
I assumed that we wanted the same behaviour as subtracting regular booleans which is
```
True - True = 0
True - False = 1
False - False = 0
False - True = -1
```
The `^` operator on np arrays does
```
np.array(True) ^ np.array(True) = False
np.array(True) ^ np.array(False) = True
np.array(False) ^ np.array(False) = False
np.array(False) ^ np.array(True) = True
```
which should do the same thing in combination with the `np.abs` and `.sum()`

skymandr: but then you don't need the `abs` any more, or?

e-larsson: Uhh true :see_no_evil:

scanomatic/image_analysis/first_pass_image.py
local-minimum: It might not be as odd as it sounds, I probably used them during the tweaking of the detection algorithm to be able to visually inspect where they ended up and they've just remained there in case they became useful another day. I'd say if they're not used we can safely remove them and should we get to rewrite grayscale detection (#33) then we probably will need to start from scratch anyhow.

local-minimum (approved):

Issue #6, opened 2021-10-19
Newer JS dependencies


local-minimum:
- [x] Use npm, don't have static dependencies
- [ ] Upgrade versions of dependencies (most important are probably `webpack` and `karma`, many others will follow with these)
- [ ] don't use `require`, use `import` (this is specifically for `d3`, see comment below)

Some of this has been done, but it would be nice to be able to use a more modern webpack, before dealing with non-dev dependencies. See comment below for some tips!

To see updateable packages you can use `$ npm outdated`. See https://www.carlrippon.com/upgrading-npm-dependencies/ for more tips.

### State of packages from 2022-01-28
```console
$ npm outdated
Package                Current   Wanted  Latest  Location                            Depended by
babel-loader             7.1.5    7.1.5   8.2.3  node_modules/babel-loader           scanomatic-standalone
babel-plugin-istanbul    4.1.6    4.1.6   6.1.1  node_modules/babel-plugin-istanbul  scanomatic-standalone
bootstrap                3.4.1    3.4.1   5.1.3  node_modules/bootstrap              scanomatic-standalone
c3                      0.4.24   0.4.24  0.7.20  node_modules/c3                     scanomatic-standalone
css-loader               5.2.7    5.2.7   6.5.1  node_modules/css-loader             scanomatic-standalone
d3                      3.5.17   3.5.17   7.3.0  node_modules/d3                     scanomatic-standalone
eslint                  7.32.0   7.32.0   8.7.0  node_modules/eslint                 scanomatic-standalone
karma-webpack           2.0.13   2.0.13   5.0.0  node_modules/karma-webpack          scanomatic-standalone
react                  16.14.0  16.14.0  17.0.2  node_modules/react                  scanomatic-standalone
react-dom              16.14.0  16.14.0  17.0.2  node_modules/react-dom              scanomatic-standalone
react-test-renderer    16.14.0  16.14.0  17.0.2  node_modules/react-test-renderer    scanomatic-standalone
style-loader             2.0.0    2.0.0   3.3.1  node_modules/style-loader           scanomatic-standalone
webpack                 4.46.0   4.46.0  5.67.0  node_modules/webpack                scanomatic-standalone
```

Note also these warnings when building, and note similar warnings when installing/updating, running `npm audit` etc!
```console
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
[...]

PR #143, opened 2022-01-27
Issue 47 validate qc part 2

skymandr:
Would have been nice to use `??`, but alas... Anyway, this very small PR makes it possible to view normalized phenotypes.

Unfortunately, D3 3.X doesn't play well with new style imports and the way we use the module, so I've changed to `require` for this. This solves the legion of console errors. (Note that D3 is now on 7.3.0, so upgrading will likely be painful...)

Part of #47 (I think I can now do QC, but won't say it's resolved because I'm not sure I've actually validated it enough, for instance I have only tried on old projects ~and there are still console errors~.)

### DIFF
--- a/scanomatic/ui_server_data/js/som/qc_normAPIHelper.js
+++ b/scanomatic/ui_server_data/js/som/qc_normAPIHelper.js
@@ -1,8 +1,9 @@
 import $ from 'jquery';
-import * as d3 from 'd3';
 
 import { getLastSegmentOfPath } from './qc_normHelper';
 import { setSharedValue } from './helpers';
+
+const d3 = require('d3/d3.js');
 
 const baseUrl = '';
 setSharedValue('baseUrl', baseUrl);
--- a/scanomatic/ui_server_data/js/som/qc_normDrawCurves.js
+++ b/scanomatic/ui_server_data/js/som/qc_normDrawCurves.js
@@ -1,5 +1,6 @@
-import * as d3 from 'd3';
 import { getExtentFromMultipleArrs, getBaseLog } from './qc_normHelper';
+
+const d3 = require('d3/d3.js');
 
 if (!d3.scanomatic) {
   d3.scanomatic = {};
@@ -74,7 +75,6 @@ d3.scanomatic.growthChart = () => {
   let generationTimeWhen;
   let generationTime;
   let growthYield;
-
 
   // local variables
   let g;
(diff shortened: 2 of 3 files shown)


scanomatic/ui_server_data/js/som/qc_normAPIHelper.js
```diff
@@ -1,9 +1,10 @@
 import $ from 'jquery';
-import * as d3 from 'd3';
 
 import { getLastSegmentOfPath } from './qc_normHelper';
 import { setSharedValue } from './helpers';
 
+const d3 = require('d3/d3.js');
```
local-minimum, line 6: What is the issue? is it that we assign things to `d3.scanomatic`? In that case, perhaps use your new get/set shared values function instead?

local-minimum: I'd really rather not have us use require

local-minimum (changes requested):

scanomatic/ui_server_data/js/som/qc_normAPIHelper.js
skymandr: That is one of the problems, but the crash happens because the code cannot address `d3.event` when in D3 is imported using modern style import. I think that is unrelated, and I can't see how using shared values for the scanomatic "plugin" would affect it. (Besides, the plugin is never accessed in the html, and the attributes set on it in one file are only accessed in that file, as far as I can tell.) That aside, doing so would also be a bit of a hacky solution, since it breaks with how you are (or were) supposed to make D3 plugins.

This stops being a problem in D3 v4, if I understand correctly, so I would say that this is transitional.

local-minimum: so plan is to do oldschool require and then make task to migrate d3 to v4 and clean up the code.

skymandr: To me that seems better than hacking around the issue. There is already a task for this, to which a note could be added https://github.com/Scan-o-Matic/scanomatic-standalone/issues/6, or do you think this requires a more specific issue?

local-minimum: just add a note

local-minimum (approved):

Issue #141, opened 2022-01-27
Validate can run Feature Extraction


e-larsson:
- [x] Feature extraction UI works
- [x] Can start feature extraction...
- [x] ...and its results can be loaded into qc (make sure it works both overwriting in an existing analysis folder and a new one)
- [ ] ~~...and extracted phenotypes are same in value to old extraction method.~~, will be fixed in #145

PR #144, opened 2022-01-28
Validate feature extraction

e-larsson:
Part of #141 

- [x] Feature extraction can be run and the results can be opened in QC.
- [x] Improved readability of the submit buttons

Part of the task was to validate that the results are the same as in the old project, and they are not. They seem to be different right from the initial project compilation step so not really sure it is due to the feature extraction.

Here is a summary of what I found

# 2 - Compile Project
With local fixture
## Produces
- `*.project.compilation`
- `*.project.compilation.instructions`

### `*.project.compilation`
- Skipped index 0 image
- Grayscale has slight differences in coordinates and decimals in values
- Orientation marks slightly different
- FixturePlateModel coordinates slightly different

## Errors
`2022-01-28 08:21:55 -- ERROR	**Compile Effector** Could not output analysis to file /somprojects/PayamMartin19Oct16Scanner2/PayamMartin19Oct16Scanner2_0000_30.2968.tiff`

# 3 - Analysis
With dynamic positioning
## Produces
- `grid_plate___*.npy`
- `grid_size___*.npy`
- `time_data.npy`
- `image_*_data.npy`

### `grid_plate___*.npy`
Same shape but slightly different values by single digits

### `image_*_data.npy`
- The file with the last index is missing, due to missing index 0 in compilation probably
- The numbers are different by about factor 2

### `time_data.npy`
- Shape is off by one, first element is missing, due to missing index 0 in compilation probably

# 4 - Feature extraction
Run with overwriting QC
## Produces
- `curves_raw.npy`
- `curves_smooth.npy`
- `normalized_phenotypes.npy`
- `phenotype_params.json`
- `phenotypes.Absolute.plate_*.csv`
- `phenotypes.extraction.instructions`
- `phenotypes_filter.npy`
- `phenotypes_meta_vector_raw.npy`
- `phenotypes_raw.npy`
- `phenotypes_reference_offsets.npy`
- `phenotype_vectors_raw.npy`
- `phenotype_times.npy`

### `curves_raw.npy` and `curves_smooth.npy`
- Shape is off by one, probably same issue as earlier
- Numbers are off by about factor 2

### `phenotypes.Absolute.plate_*.csv`
- 5 more columns in new version
- Numbers are off

### `phenotypes_filter.npy`
- Old one can't be opened

### `phenotypes_raw.npy`
- New format so difficult to compare

### `phenotype_vectors_raw.npy`
- Old one is empty

### `phenotype_times.npy`
- First element is missing

### DIFF
--- a/scanomatic/data_processing/growth_phenotypes.py
+++ b/scanomatic/data_processing/growth_phenotypes.py
@@ -363,7 +363,7 @@ def generation_time_when(
     linregress_extent: int,
     **kwargs,
 ) -> float:
-    pos = index + linregress_extent
+    pos = int(index + linregress_extent)
     if pos < 0:
         _logger.warning("No GT When because no finite slopes in data")
         return np.nan
--- a/scanomatic/data_processing/growth_phenotypes.py
+++ b/scanomatic/data_processing/growth_phenotypes.py
@@ -363,7 +363,7 @@ def generation_time_when(
     linregress_extent: int,
     **kwargs,
 ) -> float:
-    pos = int(index + linregress_extent)
+    pos = index + linregress_extent
     if pos < 0:
         _logger.warning("No GT When because no finite slopes in data")
         return np.nan
--- a/scanomatic/data_processing/phenotyper.py
+++ b/scanomatic/data_processing/phenotyper.py
@@ -1077,7 +1077,7 @@ class Phenotyper(mock_numpy_interface.NumpyArrayInterface):
                 plate_flat_regression_strided,
             ):
                 id1 = pos_index % plate.shape[1]
-                id0 = pos_index / plate.shape[1]
+                id0 = int(pos_index / plate.shape[1])
 
                 curve_data = get_preprocessed_data_for_phenotypes(
                     curve=plate[id0, id1],
--- a/scanomatic/data_processing/phenotyper.py
+++ b/scanomatic/data_processing/phenotyper.py
@@ -1021,7 +1021,7 @@ class Phenotyper(mock_numpy_interface.NumpyArrayInterface):
         all_vector_meta_phenotypes = []
 
         regression_size = self._settings.linear_regression_size
-        position_offset = (regression_size - 1) / 2
+        position_offset = (regression_size - 1) // 2
         phenotypes_count = self.get_number_of_phenotypes()
 
         total_curves = float(self.number_of_curves)
@@ -1077,7 +1077,7 @@ class Phenotyper(mock_numpy_interface.NumpyArrayInterface):
                 plate_flat_regression_strided,
             ):
                 id1 = pos_index % plate.shape[1]
-                id0 = int(pos_index / plate.shape[1])
+                id0 = pos_index // plate.shape[1]
 
                 curve_data = get_preprocessed_data_for_phenotypes(
                     curve=plate[id0, id1],
--- a/scanomatic/ui_server_data/analysis.html
+++ b/scanomatic/ui_server_data/analysis.html
@@ -130,7 +130,7 @@
             </label>
           </div>
         </div>
-        <button type="button" id="submit-button"
+        <button type="button" id="submit-button" class="submit-button"
             onclick="som.Analyse(this);">
           Run Analysis
         </button>
--- a/scanomatic/ui_server_data/compile.html
+++ b/scanomatic/ui_server_data/compile.html
@@ -70,7 +70,7 @@
           </div>
         </div>
         <div class="section">
-          <button type="button" id="submit-button"
+          <button type="button" id="submit-button" class="submit-button"
               onclick="som.Compile(this);">
             Compile Project
           </button>
--- a/scanomatic/ui_server_data/experiment.html
+++ b/scanomatic/ui_server_data/experiment.html
@@ -278,7 +278,7 @@
       </div>
 
       <div class="section">
-        <button type="button" id="submit-button"
+        <button type="button" id="submit-button" class="submit-button"
             onclick="som.StartExperiment(this)">
           Start Scanning
         </button>
--- a/scanomatic/ui_server_data/feature_extract.html
+++ b/scanomatic/ui_server_data/feature_extract.html
@@ -45,7 +45,7 @@
           </div>
         </div>
 
-        <button type="button" id="submit-button2"
+        <button type="button" id="submit-button2" class="submit-button"
             onclick="som.Extract(this);">
           Run Feature Extraction
         </button>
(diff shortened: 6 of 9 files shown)


scanomatic/data_processing/phenotyper.py
```diff
@@ -1077,7 +1077,7 @@ def _calculate_phenotypes(self):
                 plate_flat_regression_strided,
             ):
                 id1 = pos_index % plate.shape[1]
-                id0 = pos_index / plate.shape[1]
+                id0 = int(pos_index / plate.shape[1])
```
skymandr, line 1080: Shouldn't this rather be this?
```suggestion
                id0 = pos_index // plate.shape[1]
```

scanomatic/data_processing/growth_phenotypes.py
```diff
@@ -363,7 +363,7 @@ def generation_time_when(
     linregress_extent: int,
     **kwargs,
 ) -> float:
-    pos = index + linregress_extent
+    pos = int(index + linregress_extent)
```
skymandr, line 366: Why do you need `int` here when both `index` and `linregress_extent` are int?

local-minimum:
> *.project.compilation:
    Skipped index 0 image

Please make an issue out of this and place it in modernize SoM project and top priority.

local-minimum:
> Grayscale has slight differences in coordinates and decimals in values
Orientation marks slightly different
FixturePlateModel coordinates slightly different

Several detection methods are not deterministic so slight differences are expected, but if one plots the grayscales next to each other they ought to be extremely similar in shape and more or less lie on top of each other.

The original image is something like 4800x6000 pixels and the coordinates errors for orientation marks and fixture plate model should as a relative error be very small.

local-minimum:
> 2022-01-28 08:21:55 -- ERROR **Compile Effector** Could not output analysis to file /somprojects/PayamMartin19Oct16Scanner2/PayamMartin19Oct16Scanner2_0000_30.2968.tiff
This is really weird error, I suppose it output the result of the compilation somewhere? Probably make an issue for this because it would be very bad if it attempts to overwrite images.

local-minimum:
> image_*_data.npy
> 
>     The file with the last index is missing, due to missing index 0 in compilation probably
>     The numbers are different by about factor 2
> 

I agree that it's probably the missing image 0, so if you compare e.g. old index 3 with new index 2 are the numbers still that off?

local-minimum:
> phenotypes_filter.npy
> 
>     Old one can't be opened
> 

@skymandr this shouldn't be created anymore should it?

local-minimum (commented):
In general the PR looks good, but please fix/explain @skymandr code comments. Then we need to discuss and make new issues from your findings.

skymandr:
> > phenotypes_filter.npy
> > ```
> > Old one can't be opened
> > ```
> 
> @skymandr this shouldn't be created anymore should it?

The one I fixed and deprecated was `phenotype_params.npy`. Don't think this one came up as problematic in QC.

scanomatic/data_processing/growth_phenotypes.py
e-larsson: I guess it was a lazy fix but without casting I got error
```
scanomatic_1  |   File "/usr/local/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
scanomatic_1  |     self.run()
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/server/rpcjob.py", line 139, in run
scanomatic_1  |     next(effector_iterator)
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/server/phenotype_effector.py", line 134, in __next__
scanomatic_1  |     self._progress = next(self._phenotype_iterator)
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/data_processing/phenotyper.py", line 642, in iterate_extraction
scanomatic_1  |     for x in self._calculate_phenotypes():
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/data_processing/phenotyper.py", line 1110, in _calculate_phenotypes
scanomatic_1  |     phenotypes[phenotype][id0, id1] = phenotype(
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/data_processing/growth_phenotypes.py", line 640, in __call__
scanomatic_1  |     return generation_time_when(
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/data_processing/growth_phenotypes.py", line 377, in generation_time_when
scanomatic_1  |     return flat_times[pos]
scanomatic_1  | IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

skymandr: In that case it looks like the problem is further up the line. Does mypy give any hints?

scanomatic/data_processing/phenotyper.py
e-larsson: I don't mind changing, but I don't think I understand the difference in this case. If `pos_index` and `plate.shape[1]` are `int`s, is there any?

skymandr (changes requested):
To be clear, I think both the `int` casts need to be addressed.

scanomatic/data_processing/phenotyper.py
skymandr: Hmm, yeah.... As long as neither is negative, you're right. Is `pos_index` always positive (or nil)?

I think `//` is neater, but if they are always positive it isn't important.

scanomatic/data_processing/growth_phenotypes.py
e-larsson: I found the culprit and fixed it

skymandr (approved):
LGTM, but please finish the discussion with @local-minimum

skymandr:
> > > `phenotypes_filter.npy`
> > > ```
> > > Old one can't be opened
> > > ```
> > 
> > 
> > @skymandr this shouldn't be created anymore should it?
> 
> The one I fixed and deprecated was `phenotype_params.npy`. Don't think this one came up as problematic in QC. But yeah, looks like it might need conversion.

I can at least load it manually:
```python
In [3]: np.load('../Projects/PM2/PayamMartin19Oct16Scanner2/analysis/phenotypes_filter.npy', allow_pickle=True)
Out[3]: 
array([{<Phenotypes.InitialValue: 12>: array([[0, 0, 0, ..., 0, 0, 0],
...
```

PR #150, opened 2022-02-02
using default grayscale file if not found in config

BengtRydberg:
This PR allows for using a default grayscales.cfg file. Don't know if this is the best solution,
but without this scanomatic reads the file from a directory that is mounted into the container
and hence there is  no guarantee this file is present.

Anyhow, I created a new fixture,
compiled a project using this fixture,
and made a quality control that look like this:

Screenshot from 2022-02-02 20-35-56

don't really know if it looks ok, but at least the processing didn't crash.


Dockerfile
```diff
@@ -23,7 +23,9 @@ RUN echo "usb 0x4b8 0x151" >> /etc/sane.d/epson2.conf
 COPY requirements.txt /tmp/requirements.txt
 RUN pip install -r /tmp/requirements.txt
 
-COPY data/ /tmp/data/
+ENV DEFAULT_CONFIG=/tmp/data/config/
+
+COPY data/config/ ${DEFAULT_CONFIG}
```
local-minimum, line 28: this is so that the `setup.py` preforms the copy to the expected place afterwards right? But what's the point of having an environment variable?

scanomatic/data_processing/phenotyper.py
```diff
@@ -1745,7 +1745,7 @@ def _arr_tester(arr):
                 else:
                     raise ValueError("Any check failed")
             except ValueError:
-                return any(e.any() for e in arr.ravel())
+                return arr.ravel().any()
```
local-minimum, line 1748: have you validated that this change works with example in https://github.com/numpy/numpy/issues/10328

scanomatic/image_analysis/grid_array.py
```diff
@@ -31,15 +31,15 @@ def _analyse_grid_cell(
     analysis_job_model=None
 ):
     save_extra_data = grid_cell.save_extra_data
-
-    grid_cell.source = _get_image_slice(im, grid_cell).astype(np.float64)
-    if grid_cell.source is None:
+    source = _get_image_slice(im, grid_cell)
```
local-minimum, line 34: why should the `grid_cell` not be updated.

local-minimum (changes requested):

Dockerfile
BengtRydberg: the point is that this environmental variable is used in the paths.py module.

scanomatic/data_processing/phenotyper.py
BengtRydberg: it did crash it was for me, I updated this and checked that it handles the issue.

scanomatic/image_analysis/grid_array.py
BengtRydberg: fixed so grid_cell still get updated.

scanomatic/data_processing/phenotyper.py
```diff
@@ -1745,7 +1745,9 @@ def _arr_tester(arr):
                 else:
                     raise ValueError("Any check failed")
             except ValueError:
-                return any(e.any() for e in arr.ravel())
+                return any(
+                    e.any() for e in arr.ravel() if isinstance(e, np.ndarray)
```
local-minimum, line 1749: how does this work is the item is an empty dict vs a dict with something in it? In the latter case I believe it ought to return true.

Dockerfile
local-minimum: I don't think it should be fixed this way. The `setup.py` imports `setup_tools.py` script and runs `install_data_files()` with the intent of copying these files into the correct location if they are missing... something must be off there rather than solving it this way. However I believe there is a need to copy the entire `data/` into `/tmp/data`

BengtRydberg: ok, now I see the problem. install_data_files() seems to do what it is supposed to do (copy stuff to /root/.scan-o-matic). But in the docker-compose file we map a local directory at this directory, so what is there from the beginning get lost, will try to fix this in some way.

BengtRydberg: now solving this issue by modifying the entry point in the docker file so it copies required config files if missing after directories are mounted by docker-compose

local-minimum: your new settings copy thing looks nicer than my old, but perhaps remove the old one completely and cover its entire functionality (some sub directories with files) rather than have code duplication.

local-minimum:
Running `docker-compose up --build` I got:
```
scanomatic_1  |   File "/usr/local/lib/python3.9/shutil.py", line 265, in copyfile
scanomatic_1  |     with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
scanomatic_1  | IsADirectoryError: [Errno 21] Is a directory: '/tmp/data/config/ccc'
scanomatic_1  | 
scanomatic_1  | The above exception was the direct cause of the following exception:
scanomatic_1  | 
scanomatic_1  |   File "/tmp/setup_config.py", line 21, in <module>
scanomatic_1  |     setup_config()
scanomatic_1  |   File "/tmp/setup_config.py", line 17, in setup_config
scanomatic_1  |     copyfile(source_path, destination_path)
scanomatic_1  |   File "/usr/local/lib/python3.9/shutil.py", line 293, in copyfile
scanomatic_1  |     raise FileNotFoundError(f'Directory does not exist: {dst}') from e
scanomatic_1  | FileNotFoundError: Directory does not exist: root/.scan-o-matic/config/ccc
scanomatic-standalone_scanomatic_1 exited with code 1
```

Probably missing some making of directories of the subdirectories.

local-minimum (commented):
Besides that it looks good now

local-minimum (approved):
I think the PR looks good (and should be merged as is), but I don't think it solves the validate fixture task because I simply cannot get any grayscale detection with the correct grayscale (silverfast) and the detection with the wrong one (kodak) looks funky

BengtRydberg:
> I think the PR looks good (and should be merged as is), but I don't think it solves the validate fixture task because I simply cannot get any grayscale detection with the correct grayscale (silverfast) and the detection with the wrong one (kodak) looks funky
> 
> image

I have managed to get detection with silverfast, but I had to try a number of times to get it to work...

local-minimum:
Can you screenshot how the curve of the detection looks?

BengtRydberg:
> Running `docker-compose up --build` I got:
> 
> ```
> scanomatic_1  | Traceback (most recent call last):
> scanomatic_1  |   File "/usr/local/lib/python3.9/shutil.py", line 265, in copyfile
> scanomatic_1  |     with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
> scanomatic_1  | IsADirectoryError: [Errno 21] Is a directory: '/tmp/data/config/ccc'
> scanomatic_1  | 
> scanomatic_1  | The above exception was the direct cause of the following exception:
> scanomatic_1  | 
> scanomatic_1  | Traceback (most recent call last):
> scanomatic_1  |   File "/tmp/setup_config.py", line 21, in <module>
> scanomatic_1  |     setup_config()
> scanomatic_1  |   File "/tmp/setup_config.py", line 17, in setup_config
> scanomatic_1  |     copyfile(source_path, destination_path)
> scanomatic_1  |   File "/usr/local/lib/python3.9/shutil.py", line 293, in copyfile
> scanomatic_1  |     raise FileNotFoundError(f'Directory does not exist: {dst}') from e
> scanomatic_1  | FileNotFoundError: Directory does not exist: root/.scan-o-matic/config/ccc
> scanomatic-standalone_scanomatic_1 exited with code 1
> ```
> 
> Probably missing some making of directories of the subdirectories.

I changed the script so it only copies files. 




> Can you screenshot how the curve of the detection looks?
Screenshot from 2022-02-04 10-37-21

local-minimum:
Great that looks like a proper GS too. Lets say it resolves the thing then (you'll have to re-add it)

PR #155, opened 2022-02-08
regridding ui

### LINKED ISSUE #153: Regridding UI doesn't work
# 🐛 Description

Given that there's an analysis it should be possible to to a regridding

# 🔢 To reproduce

1. Create an analysis 
2. Go back to analysis step
3. Fill out the form
4. The checkbox to enable regridding does not become enabled.

# ✌🏽 Expected behaviour

Should be possible to enable.

BengtRydberg:
This small PR deals with getting the regridding UI to show up / work on the analysis page.
The screenshot below shows that this PR make it possible to enable the regridding:
Screenshot from 2022-02-09 12-43-44

I first made an analysis without regridding, and here is a quality control from this:

Screenshot from 2022-02-09 14-19-17

I then made an analysis with regridding (with regridding setting as in the first figure), and the quality control looks like this:

Screenshot from 2022-02-09 14-18-13

the two figures clearly differ but do not know if this looks ok.

Resolves #153

### DIFF
--- a/scanomatic/ui_server_data/analysis.html
+++ b/scanomatic/ui_server_data/analysis.html
@@ -5,6 +5,8 @@
   <script src="/js/somlib/som.js"></script>
   <script>
     som.setSharedValue("currentFixtureId", "#current-fixture");
+    som.setSharedValue("newFixtureDataId", "#new-fixture-data");
+    som.setSharedValue("selectedFixtureDivId", "#selected-fixture");
   </script>
   <link rel="stylesheet" type="text/css" href="style/main.css?ver=_-_VERSIONTAG_-_">
   <link rel="stylesheet" type="text/css" href="style/analysis.css?ver=_-_VERSIONTAG_-_">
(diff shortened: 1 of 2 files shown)


local-minimum:
Have you validated that if you move the grid on one of the plates (even though it is actually correct), that when doing the qc you can see it actually moved too?

local-minimum:
But to resolve the issue I'd like to at least know that reasonable request is sent to the backend and preferably that it worked or created a new issue if it didn't work.

BengtRydberg:
> But to resolve the issue I'd like to at least know that reasonable request is sent to the backend and preferably that it worked or created a new issue if it didn't work.

I have now made a regridding analysis, see figures above.

local-minimum:
If you click the show grid image button you should be able to see the new grid image and that ought to have a grid offset from what it originally correctly was in accordance with your applied offset. Or at least something close to applying the requested offset as things aren't deterministic. But I think it looks good enough to close the issue for now.

local-minimum (approved):

PR #157, opened 2022-02-11
Split status

joakimmoller:
Makes server status work, reduces complexity in "/api/status" path

### DIFF
--- a/scanomatic/io/rpc_client.py
+++ b/scanomatic/io/rpc_client.py
@@ -83,8 +83,9 @@ class _ClientProxy:
             Popen(["scan-o-matic_server"])
         else:
             self._logger.warning(
-                "Can't launch because server is %s",
-                ['not local', 'online'][self.online]
+                "Can't launch because server is {0}".format(
+                    ['not local', 'online'][self.online],
+                )
             )
 
     def __getattr__(self, key):
@@ -108,7 +109,7 @@ class _ClientProxy:
             self._logger.info("No client active")
         else:
             address = "{0}:{1}/".format(self._host, self._port)
-            self._logger.info("Communicates with '%s'", address)
+            self._logger.info("Communicates with '{0}'".format(address))
             self._client = xmlrpc.client.ServerProxy(address)
 
     def user_id_decorator(self, f: Callable):
@@ -125,10 +126,11 @@ class _ClientProxy:
                 return f(*args, **kwargs)
             except xmlrpc.client.Fault:
                 self._logger.critical(
-                    "Failed to communicate with %s(%s, **%s)",
-                    f.__name__,
-                    args,
-                    kwargs,
+                    "Failed to communicate with {}({}, **{})".format(
+                        f.__name__,
+                        args,
+                        kwargs,
+                    ),
                 )
                 raise
 
@@ -150,11 +152,10 @@ class _ClientProxy:
                     and not (self._user_id is None and v in self._admin_methods)
                 )
             except socket.error:
-                self._logger.warning(
-                    "Connection Refused for '%s:%i'",
+                self._logger.warning("Connection Refused for '{0}:{1}'".format(
                     self.host,
                     self.port,
-                )
+                ))
                 return ("launch_local",)
 
         return ret
--- a/scanomatic/io/rpc_client.py
+++ b/scanomatic/io/rpc_client.py
@@ -83,9 +83,8 @@ class _ClientProxy:
             Popen(["scan-o-matic_server"])
         else:
             self._logger.warning(
-                "Can't launch because server is {0}".format(
-                    ['not local', 'online'][self.online],
-                )
+                "Can't launch because server is %s",
+                ['not local', 'online'][self.online]
             )
 
     def __getattr__(self, key):
@@ -98,8 +97,7 @@ class _ClientProxy:
                 ]
             )
             return m
-        else:
-            raise AttributeError(f"Client doesn't support attribute {key}")
+        raise AttributeError(f"Client doesn't support attribute {key}")
 
     def __dir__(self):
         return list(self._allowed_methods())
@@ -110,7 +108,7 @@ class _ClientProxy:
             self._logger.info("No client active")
         else:
             address = "{0}:{1}/".format(self._host, self._port)
-            self._logger.info("Communicates with '{0}'".format(address))
+            self._logger.info("Communicates with '%s'", address)
             self._client = xmlrpc.client.ServerProxy(address)
 
     def user_id_decorator(self, f: Callable):
@@ -127,11 +125,10 @@ class _ClientProxy:
                 return f(*args, **kwargs)
             except xmlrpc.client.Fault:
                 self._logger.critical(
-                    "Failed to communicate with {}({}, **{})".format(
-                        f.__name__,
-                        args,
-                        kwargs,
-                    ),
+                    "Failed to communicate with %s(%s, **%s)",
+                    f.__name__,
+                    args,
+                    kwargs,
                 )
                 raise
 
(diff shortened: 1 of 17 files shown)


BengtRydberg (approved):
LGTM

scanomatic/io/rpc_client.py
```diff
@@ -83,9 +83,8 @@ def launch_local(self) -> None:
             Popen(["scan-o-matic_server"])
         else:
             self._logger.warning(
-                "Can't launch because server is {0}".format(
-                    ['not local', 'online'][self.online],
-                )
+                "Can't launch because server is %s",
```
local-minimum, line 86: don't introduce `%` formatting rather use `f"..."`

scanomatic/io/rpc_client.py
```diff
@@ -127,11 +125,10 @@ def _wrapped(*args, **kwargs):
                 return f(*args, **kwargs)
             except xmlrpc.client.Fault:
                 self._logger.critical(
-                    "Failed to communicate with {}({}, **{})".format(
-                        f.__name__,
-                        args,
-                        kwargs,
-                    ),
+                    "Failed to communicate with %s(%s, **%s)",
```
local-minimum, line 128: same

scanomatic/io/rpc_client.py
```diff
@@ -153,10 +150,11 @@ def _allowed_methods(self):
                     and not (self._user_id is None and v in self._admin_methods)
                 )
             except socket.error:
-                self._logger.warning("Connection Refused for '{0}:{1}'".format(
+                self._logger.warning(
+                    "Connection Refused for '%s:%i'",
```
local-minimum, line 154: same

scanomatic/ui_server/status_api.py
```diff
@@ -1,63 +1,72 @@
...
-                return jsonify(
-                    scanners={
-                        s['socket']: s['scanner_name']
-                        for s in rpc_client.get_scanner_status()
-                        if 'owner' not in s or not s['owner']
-                    },
+@blueprint.before_request
+def get_rpc():
+    if 'rpc' not in g:
+        g.rpc = get_client()  # noqa
```
local-minimum, line 13: what is the noqa for here, and if it really is needed specify the exact rule that should be disabled never use just noqa

local-minimum: if this is a preformance optimization it seems more reasonable to put an lru_cache wrapper around `get_client` rather than putting it on the global `g` as I assume it won't be typed properly when used.

scanomatic/ui_server/ui_server.py
```diff
@@ -46,14 +46,15 @@ def launch_server(host, port, debug):
...
+    if not rpc_client.online:
+        if rpc_client.local:
+            _LOGGER.warning(
+                "No local RPC Server detected launching local instance.",
+            )
+            rpc_client.launch_local()
+    else:
         _LOGGER.error(
-            f"Can't reach RPC Server at {rpc_client.host}:{rpc_client.port}",
+            "Can't reach RPC Server at %s:%i", rpc_client.host, rpc_client.port
```
local-minimum, line 57: same

scanomatic/ui_server/ui_server.py
```diff
@@ -63,10 +64,12 @@ def launch_server(host, port, debug):
 
     _URL = f"http://{host}:{port}"
     _LOGGER.info(
-        f"Requested to launch UI-server at {_URL} being debug={debug}",
+        "Requested to launch UI-server at %s being debug=%s",
```
local-minimum, line 67: same

scanomatic/ui_server/ui_server.py
```diff
@@ -138,7 +143,7 @@ def _js_external(group, js):
 
 def launch_webbrowser(delay=0.0) -> None:
     if delay:
-        _LOGGER.info(f"Will open webbrowser in {delay} s")
+        _LOGGER.info("Will open webbrowser in %f s", delay)
```
local-minimum, line 146: same

scripts/scan-o-matic
```diff
@@ -71,7 +71,7 @@ if __name__ == "__main__":
         default=False,
         action='store_true',
         help="Run in debug-mode. This makes the app vunerable and insecure" +
-        " and should be used behind firewall"  +
+        " and should be used behind firewall" +
```
local-minimum, line 74: next line should probably have a leading space in the string

scripts/scan-o-matic
```diff
@@ -98,7 +98,9 @@ if __name__ == "__main__":
         )
         for proc in procs:
             _logger.info(
-                f"Killing process '{proc.name()}' with pid {proc.pid}",
+                "Killing process '%s' with pid %i",
```
local-minimum, line 101: same

local-minimum (changes requested):

scanomatic/io/rpc_client.py
joakimmoller: It's not so much a preference of type of string formatting interpolation to use, it's a matter of using the features that logging offers. You don't want string interpolation to happen if the message isn't going to be published.

scripts/scan-o-matic
joakimmoller: I hade no intention to change the string at all just a automatic white space removal. String changed.

scanomatic/ui_server/status_api.py
joakimmoller: It's for muting pylint, removed.

scanomatic/io/rpc_client.py
local-minimum: I understand the argument but I don't agree, I think the readability of `%` formatting use so much worse than f-strings, which is a general tendency not only in python to format that way that I think there's a strong argument to do string formatting even if it is useless when not output. That sound like chasing performance that shouldn't be that important

scanomatic/ui_server/status_api.py
```diff
@@ -1,63 +1,72 @@
...
+@blueprint.before_request
+def get_rpc():
+    if 'rpc' not in g:
+        g.rpc = get_client()
+
+
+@blueprint.route("/server")
+def status_server():
+    if g.rpc.online:
+        return jsonify(**g.rpc.get_status())
```
local-minimum, line 19: is `g.rpc` correctly typed, if not, then I don't think it's sufficently beneficial to do get_client() in the before_request.

joakimmoller: Don't understand what you what me to do? call get_client() in each route?

scanomatic/io/rpc_client.py
joakimmoller: I don't find it very hard to read the %-formatting, not using a module as intended disturbs me more.

local-minimum: I'm sorry but I don't agree it is much harder to follow for me, and if we should do such a change we should do it after agreeing on it and reworking it across all files.

scanomatic/ui_server/status_api.py
local-minimum: yes rather that if `rpc` isn't typed on `g.rpc`. `get_client` is meant to be caching so should not be costly and if it was a cost issue it's just a matter of slapping a `lru_cache` on it

scanomatic/io/rpc_client.py
joakimmoller: OK, I'll back out changes in logging.

scanomatic/ui_server/status_api.py
local-minimum: or alternatively should `g.rpc` not be propertly typehinted it can be cast each use.

local-minimum (approved):

PR #159, opened 2022-02-14
Fix scanning

### LINKED ISSUE #45: Validate that can run scan project
# 📋 Info

Run a project for a couple of scans (10+)

# 🏁 DoD

- [ ] Scanning works
- [ ] Makes a copilation and analysis

e-larsson:
Resolves #45 

- Scanning works, tested all different parameters
- A compilation is created and subsequent steps are run

Not very happy with the solution for parsing Enums in the abstract factory since it requires all enums that are used in that way to have the `from_name` property. If you have another simple solution that would be appreciated


scanomatic/io/app_config.py
```diff
@@ -249,12 +249,11 @@ def _generate_admin_uuid(self) -> Optional[str]:
...
-            dump(
-                self._settings,
-                self._paths.config_main_app,
-                merge=True,
-            )
+        dump(
+            self._settings,
+            self._paths.config_main_app,
+            merge=True,
+        )
```
skymandr, line 256: Why no `validate`?

skymandr (dismissed):
Just a quick question so far, but I want to look at this. Also, be aware that you are introducing new type-errors with this! It would be nicer to see the number going down... ;)

docker-compose.yml
```diff
@@ -5,6 +5,8 @@ services:
     privileged: true
     build: .
     ports: ["5000:5000"]
+    environment:
+      - LOGGING_LEVEL=10
```
local-minimum, line 9: it's not gonna spew out tons of logs or is it a reasonable amount?

scanomatic/generics/abstract_model_factory.py
```diff
@@ -347,14 +354,16 @@ def _enforce_other(dtype, obj):
...
+                    settings[key] = _get_enum(dtype, settings[key])
                 else:
                     settings[key] = _enforce_other(dtype, settings[key])
             # else it is already correct type
 
     @classmethod
     def to_dict(cls, model) -> dict:
         model_as_dict = dict(**model)
-        for k in model_as_dict.keys():
+        for k in list(model_as_dict.keys()):
```
local-minimum, line 366: this list seems to not do anything?

scanomatic/generics/abstract_model_factory.py
```diff
@@ -241,6 +242,12 @@ def _enforce_model(factory, obj):
                 else:
                     break
 
+        def _get_enum(enum: EnumMeta, name: str) -> Enum:
```
local-minimum, line 245: why is this needed? loaded models should already have enums as enums when it reaches the factory.

scanomatic/io/app_config.py
local-minimum: please re-add validation, though it should probably not be `self.validate` but from the `validate.py`

scanomatic/generics/abstract_model_factory.py
local-minimum: The `object_hook` in `jsonizer.py` should have loaded all enums as enums.

docker-compose.yml
skymandr, line 9: This enables debug-logging by default... I don't think that's a good idea. `20` would be a more appropriate default.

scanomatic/io/scanner_manager.py
```diff
@@ -75,7 +75,9 @@ def _initiate_scanners(self) -> dict[int, ScannerModel]:
         scanner_configs = load(self._paths.config_scanners)
         for scanner in cast(
             list[ScannerModel],
-            [] if scanner_configs is None else scanner_configs,
+            [] if scanner_configs is None else
```
local-minimum, line 78: two if else in one line feels very iffy.  please at least add parenthesis but preferably rewrite it

scanomatic/models/analysis_model.py
```diff
@@ -23,12 +23,24 @@ class IMAGE_ROTATIONS(Enum):
     Portrait = 1
     Unknown = 2
 
+    @classmethod
+    def from_name(cls, raw_string: str) -> "IMAGE_ROTATIONS":
```
local-minimum, line 27: why not just `IMAGE_ROTATIONS['Unknown']`?

local-minimum: or are some recorded in lower-case? and if this function really is needed then it needs to raise value error if no member found

local-minimum: but I really doubt we need it since we could load all serialized states before with jsonizer.

scanomatic/server/interface_builder.py
```diff
@@ -215,20 +215,22 @@ def _server_get_status(self, user_id=None):
...
-                    for scanner_model in _SOM_SERVER.scanner_manager.status
-                ],
-                key=lambda x: x['socket'],
-            ))
+        self.logger.debug(
+            f"Scanner manager status: {_SOM_SERVER.scanner_manager.status}"
+        )
+        return sanitize_communication(sorted(
+            [
+                ScannerFactory.to_dict(scanner_model)
```
local-minimum, line 229: this might be the problem, we should probably not use or have the `to_dict` left unless it does something very magical and instead use `jsonizer.dumps`

scanomatic/server/scanning_effector.py
```diff
@@ -975,7 +977,7 @@ def _scan_thread(self):
         )
 
     def _setup_directory(self):
-        os.makedirs(self._project_directory)
+        os.makedirs(self._project_directory, exist_ok=True)
```
local-minimum, line 980: this is actually not really that good, I think there's a point to make that when starting a project it must be a new directory

scanomatic/ui_server/status_api.py
```diff
@@ -52,7 +52,7 @@ def status_scanners(status_query=None):
                 scanners={
                     s['socket']: s['scanner_name']
                     for s in rpc_client.get_scanner_status()
-                    if 'owner' not in s or not s['owner']
+                    if s.get('owner', {}).get('pid', 0) == 0
```
local-minimum, line 55: what is the logic for this here, it looks odd

local-minimum (changes requested):

scanomatic/ui_server_data/templates/settings.html
```diff
@@ -77,7 +77,7 @@ <h2>Other</h2>
       </div>
 
       <div class="section">
-        <button type="button" id="submit-button" onclick="som.UpdateSettings(this, false);">Save (will take effect next reboot of Scan-o-Matic & Scan-o-Matic server)</button>
+        <button type="button" id="submit-button" class="submit-button" onclick="som.UpdateSettings(this, false);">Save (will take effect next reboot of Scan-o-Matic & Scan-o-Matic server)</button>
```
skymandr, line 80:
```suggestion
        <button type="button" id="submit-button" class="submit-button" onclick="som.UpdateSettings(this, false);">
          Save (will take effect next reboot of Scan-o-Matic & Scan-o-Matic server)
        </button>
```

skymandr (commented):
I'm afraid I cannot help with the enums. I have no objections (but a few small comments), but please wait for @local-minimum to have a look.

docker-compose.yml
skymandr: `10` enables debug. In my opinion `20` (info) would be a better default.

scanomatic/io/app_config.py
e-larsson: I'm confused :confused: . I was sure this change was done by @joakimmoller but it looks like it's mine. I'll reintroduce it

scanomatic/generics/abstract_model_factory.py
e-larsson: Without it I get
```
canomatic_1  | 2022-02-15 08:19:49 -- ERROR    **Scan-o-Matic UI** Exception on /api/status/scanners/free [GET]
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 2073, in wsgi_app
scanomatic_1  |     response = self.full_dispatch_request()
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1518, in full_dispatch_request
scanomatic_1  |     rv = self.handle_user_exception(e)
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/flask_restful/__init__.py", line 271, in error_router
scanomatic_1  |     return original_handler(e)
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1516, in full_dispatch_request
scanomatic_1  |     rv = self.dispatch_request()
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1502, in dispatch_request
scanomatic_1  |     return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/ui_server/status_api.py", line 54, in status_scanners
scanomatic_1  |     for s in rpc_client.get_scanner_status()
scanomatic_1  |   File "/usr/local/lib/python3.9/site-packages/scanomatic/io/rpc_client.py", line 126, in _wrapped
scanomatic_1  |     return f(*args, **kwargs)
scanomatic_1  |   File "/usr/local/lib/python3.9/xmlrpc/client.py", line 1122, in __call__
scanomatic_1  |     return self.__send(self.__name, args)
scanomatic_1  |   File "/usr/local/lib/python3.9/xmlrpc/client.py", line 1464, in __request
scanomatic_1  |     response = self.__transport.request(
scanomatic_1  |   File "/usr/local/lib/python3.9/xmlrpc/client.py", line 1166, in request
scanomatic_1  |     return self.single_request(host, handler, request_body, verbose)
scanomatic_1  |   File "/usr/local/lib/python3.9/xmlrpc/client.py", line 1182, in single_request
scanomatic_1  |     return self.parse_response(resp)
scanomatic_1  |   File "/usr/local/lib/python3.9/xmlrpc/client.py", line 1354, in parse_response
scanomatic_1  |     return u.close()
scanomatic_1  |   File "/usr/local/lib/python3.9/xmlrpc/client.py", line 668, in close
scanomatic_1  |     raise Fault(**self._stack[0])
scanomatic_1  | xmlrpc.client.Fault: <Fault 1: "<class 'RuntimeError'>:dictionary changed size during iteration">
```

scanomatic/server/scanning_effector.py
joakimmoller: There is something else creating that directory, it always exists and prohibits scanning. One could probably remove _setup_directory...

scanomatic/ui_server/status_api.py
joakimmoller: The real problem is that the "new model" initiates a free scanner object with  {'owner': 'id':0}  but the old parts of scanomatic considers an object with out 'owner' as a free scanner.

scanomatic/server/scanning_effector.py
local-minimum: if it always exists then rather remove it if it isn't needed.

scanomatic/ui_server/status_api.py
local-minimum: does that mean that the `pid` no longer is optional / can that be updated on the model typing?

scanomatic/io/app_config.py
local-minimum: I think it might have been me that broke this when updating models but any model derived from `generics.model.Model` (or what the path is should be possible to validate with `validate.py` without any fuss.

e-larsson: Looking at this more closely I'm not sure it's wrong.

`self.validate()` was removed previously so if validation should occur it should def be from the `validate.py`

But the only place `save_current_settings()` is called is here where the settings are already validated before saving, https://github.com/Scan-o-Matic/scanomatic-standalone/blob/a95d1082989c2620eac64ce3757ae063f1bacbe9/scanomatic/ui_server/settings_api.py#L42

Importing the `validate()` function also causes a big mess of circular imports.

docker-compose.yml
e-larsson: Changed to 10

scanomatic/generics/abstract_model_factory.py
e-larsson: I'm super confused about this. I've spent I don't know how much time trying to solve this enum parsing issue. But now it just works without it. :confused:
Anyway, I removed it... :facepalm:

scanomatic/io/scanner_manager.py
e-larsson: Refactored this a bit

scanomatic/server/interface_builder.py
e-larsson: There is some magic going on it seems. `jsonizer.dumps` causes scanners to not show up

skymandr: The `dumps` doesn't make an identical structure to `to_dict`, keeping track of types etc, so probably rather than magic, it is a matter of different expectations?

docker-compose.yml
local-minimum: you mean 20

scanomatic/io/app_config.py
local-minimum: I see, in that case we don't need to validate it twice. Perhaps just add a docstring to the `save_current_settings` that it should be validated beforehand and also perhaps it would be nicer to not save `self._settings` but `self.application_settings` since that is what is validated.

scanomatic/io/scanner_manager.py
```diff
@@ -73,10 +73,12 @@ def _initiate_scanners(self) -> dict[int, ScannerModel]:
...
-        for scanner in cast(
-            list[ScannerModel],
-            [] if scanner_configs is None else scanner_configs,
-        ):
+        if scanner_configs is None:
+            scanner_configs = []
+        elif not isinstance(scanner_configs, list):
+            scanner_configs = [scanner_configs]
+
+        for scanner in cast(list[ScannerModel], scanner_configs):
```
local-minimum, line 81: much nicer

scanomatic/server/interface_builder.py
local-minimum: ah yes this is for rpc communication and that shouldn't need preserializing everything. But now looks nice I think

scanomatic/generics/abstract_model_factory.py
local-minimum: So it is some times.

local-minimum (commented):
Mostly want answers to the two outstanding questions if the makedirs could be removed then and if the `pid` as the correct typehint now if it always has a value

scanomatic/server/scanning_effector.py
joakimmoller: The mechanism that creates the project dir is the logger. So it's alright to remove the `_setup_directory` function, but it really not a good idea. When/if we replace/change the logger noone creates the project dir. It is probably safe to keep `exists_ok=True` since the experiment-page won't let you to continue if the dir already exists.

scanomatic/io/app_config.py
e-larsson: Added docstring and changed `self._settings -> self.application_settings`

local-minimum (approved):

